PR #71, opened 2014-06-26
Add support for vmware_fusion

josegonzalez:


josegonzalez:
Closing since it's not going to be merged.

Just a note: might be helpful to contributors if you commented as to why you won't fix an issue as opposed to just labeling it as such.

PR #78, opened 2014-07-09
Create run.sh

cauerego:
README.md would obviously need to change one line to:

` wget -qO- https://raw.githubusercontent.com/scrapinghub/portia/master/install.sh | bash `

So it would the last ` install.sh ` line which has a similar `wget` for `run.sh` - not sure on a better way to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Go inside:
 
 Then simply run:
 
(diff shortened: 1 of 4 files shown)


ruairif:
It's now possible to run Portia using Docker, Vagrant or installing it locally.
While this would make it slightly easier it also might not work in all cases (missing system dependencies for lxml for example).

cauerego:
Cool, and fair enough. Docker rocks. :)

PR #79, opened 2014-07-23
Mining listing data

tpeng:
* UI:
add listingData checkbox in annotation window, so annotation can be marked as `listingData`, those annotations will be handled by `MDRExtractor`.

* backend:
add SlybotIBL and enable `MDRExtractor`

see also https://github.com/scrapy/scrapely/pull/58


ruairif:
Fixed by using Containers to define page regions

PR #89, opened 2014-08-19
A bit more doc and a missing dependency

boogheta:
While trying to install portia I had to search to find out about these, so I hope this will save some time to others


ruairif:
`service_identity` is now included in the requirements.
The current install instructions instruct users to install Portia in such a way that `portiacrawl` ends up in the path.

PR #135, opened 2015-01-26
be more docker layer-cache friendly

dangra:


dangra:
I used a different requirements.txt file with pinned dependencies versions so it only expire the "pip install -r requirements.txt" layer when a version or new dep is added to this file.

dangra:
Also based the image on ubuntu:12.04 as `base:12.04` is an internal image not available to open public.

dangra:
@ruairif this is ready to merge.

PR #140, opened 2015-02-13
Fixed issue with externalized templates

mbrunthaler:
portiacrawl did not load externalized templates from the spec file. This
change loads and adds external templates from the spec directory.


ruairif:
Looks good. Thanks for the fix

PR #144, opened 2015-02-25
Added dependencies in readme

flip111:


README.md
```diff
@@ -176,12 +176,17 @@ Instead we need to modify layout A's template, and mark the ``description`` anno
 Running Portia
 --------------
 
-Checkout the repository:
+### Dependencies ###
```
ruairif: This should be a sub section of `Running Portia Locally`.

README.md
```diff
@@ -176,12 +176,17 @@ Instead we need to modify layout A's template, and mark the ``description`` anno
...
+You might need to install the following packages
```
ruairif: Mention this is for Ubuntu or Debian. OSX and Windows users can install the packages as binaries from pip.

"When running Portia on Ubuntu or Debian systems you may need to install the following dependencies:"

README.md
```diff
@@ -176,12 +176,17 @@ Instead we need to modify layout A's template, and mark the ``description`` anno
...
 
-    git clone https://github.com/scrapinghub/portia
+    sudo apt-get install libxml2-dev libxslt-dev python-dev zlib1g-dev libffi-dev libssl-dev
```
ruairif: I think build-essential should be here too.

Are you sure about the zlib dependency? Is so do you know which library requires it?

flip111: build-essential was already installed on my system so i didn't add it.

zlib was needed i think by numpy. If you really want to know i will start from scratch and show you the log file. Let me know if this is needed.

PR #164, opened 2015-03-12
Ignoring sub-elements now works and can be viewed in context to the annotation which is its parent element

jskrill:


slyd/app/components/annotations-plugin/component.js
```diff
@@ -132,10 +132,10 @@ export default Ember.Component.extend({
         },
 
         ignoreBeneath: function(_, value, index) {
-            var ignore = this.get('ignores').get(index),
+            var ignore = this.get('ignores').objectAt(index),
                 ignoreData = this.get('alldata').findBy('tagid', ignore.tagid);
-            ignore.ignoreBeneath = value;
-            ignoreData.ignoreBeneath = value;
+            ignore.set('ignoreBeneath', value);
+            ignoreData['ignore_beneath'] = value;
```
ruairif: Add `.set` here too

slyd/.gitignore
```diff
@@ -20,3 +20,6 @@ data/*
 # npm files
 node_modules/*
 npm-debug.log
+
+#ember build
+dist/*
```
ruairif: We are shipping the compiled assets to users in the dist folder.

The command below is used to build the updated js:

    `ember build -e production`

slyd/slyd/plugins/scrapely_annotations/annotations.py
```diff
@@ -13,6 +13,23 @@
...
+            extracts: [
+                {
+                    annotatations: {"content": "Title"},
+                    id: "id-string",
+                    required: [],
+                    tagid: 12,
+                    variant: 0
+                    ignore: True,
+                    ignore_beneath: true,
+                    insert_after: true,
```
ruairif: Capital `T` to make it match the python `True` used above.

slyd/app/components/annotations-plugin/component.js
```diff
@@ -584,6 +593,28 @@ export default Ember.Component.extend({
         }
     },
 
+    setIgnores: function() {
```
ruairif: You can make this into a shared variable stored in `pluginState` to match the behaviour of `maxVariant` and `maxSticky`

slyd/app/components/annotations-plugin/component.js
jskrill: .set isn't defined here because the elements of alldata aren't ember objects but regular javascript objects. I tried to find where alldata was being set and saw that it was activeExtractionTool.data.extracts which is set as this.get('extractionTools.' + tool_name) in controllers/template.js and that finally seems to map to model. It seems model is set before enableExtractionTool and thus is never set as this.get('annotationsStore').findAll() which would be ember objects. I couldn't find where model could be set before that function call. Forcing it to set alldata to this.get('annotationsStore').findAll() breaks quite a bit of code.

slyd/app/components/annotations-plugin/component.js
jskrill: I'm not sure what you mean? Should the array of ignores be accessed through the this.get('pluginState.ignores') as opposed to this.get('ignores')?

ruairif: That's correct.

It should be set with:

    this.set('pluginState.ignores', ignores);
    this.updateData('pluginState');

This is to allow it to better handle updates to ember in the future.

slyd/app/components/annotations-plugin/component.js
ruairif: You're right. I didn't realise that it was `ignoreData` not `ignore`.

There are some workarounds there to support an old data format. I'll be cleaning that up a bit for Ember 1.12 or 1.13 depending on the changes that they have planned.

slyd/app/components/annotations-plugin/component.js
```diff
@@ -850,12 +887,13 @@ export default Ember.Component.extend({
         this.set('data', null);
         this.set('mappedElement', null);
         this.set('mappedDOMElement', null);
-        this.set('ignores', []);
+        this.set('pluginState.ignores', []);
```
ruairif: This is resetting the global plugin state. I don't think this needs to be reset and the pluginstate object doesn't need to be updated.

PR #169, opened 2015-03-19
itemDefinitions is now updated before calling wrapItem

jskrill:
definitionItems wasn't being updated so when the extracted items were being iterated over to be returned in the fields function of the ExtractedItem model new fields would the conditional in
```
var fieldDefinition = this.get('definition.fields').findBy('name', key);
if (fieldDefinition) {
    fields.pushObject(ExtractedField.create(
        { name: key, type: fieldDefinition.get('type'), value: item[key] }));
    }
}.bind(this));
```
Alot of files which were changed in dist were uploaded with this push. Should they be merged into the project?


slyd/app/controllers/spider.js
```diff
@@ -226,8 +226,13 @@ export default BaseController.extend({
     },
 
     updateExtractedItems: function(items) {
-        var extractedItems = items.map(this.wrapItem, this);
-        this.set('extractedItems', extractedItems);
+        this.get('slyd')
+        .loadItems()
+        .then(function(updatedItemDefinitions) {
```
ruairif: This will fetch `items` for every page load request.

It would be better to have the spider controller and the template controller share a base controller (as controllers are singletons) and have them share items through that.

ruairif:
The extra files are from the debug build of Portia assets. If you want to include assets be sure to build them with `ember build -e production`

ruairif:
If you look [here](https://github.com/scrapinghub/portia/blob/master/slyd/app/initializers/project-models.js) you can see that each controller has a `project_models` attribute that is shared. It should be possible to use that to avoid additional requests to the server (which is a problem that it has at the moment anyway).

Issue #138, opened 2015-02-11
Exception when adding invalid domain


ruairif:
Steps
- Open a project in Portia
- Add new spider for domain "test"
- Open the start URL

You will get the following exception. We should style it better. Or avoid adding spiders with bad domains.
"unexpected error response: [Failure instance: Traceback (failure with no frames): : DNS lookup failed: address 'test3' not found: [Errno -2] Name or service not known. ]"

PR #174, opened 2015-03-26
Spiders of inaccessible domains can't be created

jskrill:
Fix for #138 

I opted to disallow the creation of spiders of invalid domains. The only real way I could think of testing this was using to make a request to the url. fetchDocument returns quite a bit more then that however and upon spider creation it would be called immediately again so I was considering storing the return of fetchDocument and then making a check for it when the page navigates to the spiders route. This perhaps could be tied into general page caching functionality.


ruairif:
Looks good so far.

To avoid the extra page load you could store the response in the `project_models` object (`project_models.newSpiderPage`?). Then when you load the `spider` controller if you hook into [here](https://github.com/scrapinghub/portia/blob/master/slyd/app/controllers/spider.js#L605) and move all of [this](https://github.com/scrapinghub/portia/blob/master/slyd/app/controllers/spider.js#L247-L268) to a separate function then it should be doable.
Remember to set the page in the global model back to null afterwards to be safe.

jskrill:
[This](https://github.com/scrapinghub/portia/blob/master/slyd/app/controllers/spider.js#L603-610) block of code never actually ran because this.get('controllers.application.siteWizard'); was being set to null in addSpider. When it did run and I made my changes, the page was fetching anyway because of [this](https://github.com/scrapinghub/portia/blob/master/slyd/app/controllers/spider/index.js#L15-21) block of code so I made it so that would only run if project_models.newSpiderPage was null.

ruairif:
Thanks for the help.

PR #186, opened 2015-04-10
Slybot version number

agusc:


pablohoffman:
no objections, @dangra ?

agusc:
Probably that script isn't necessary anymore and we can leave the default build command, removing slybot/Makefile.buildbot from repo. Default build command adds by default the hash commit into version.
I preferred to just add it there because I wasn't 100% sure if that custom build command (bin/makedeb) is still being used.

dangra:
no objections except we can try first to remove Makefile.buildbot and rely on default action.

PR #205, opened 2015-06-04
Added documentation for field types

rdowinton:
I've added documentation for field types in Portia. There have been questions from a few customers about what specifically safe html does, and it's useful to be more clear about how each field type works on a technical level

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,45 +89,7 @@ Setting a field's type will ensure the field will only match certain types of da
     </tbody>
 </table>
 
-The ``safe html`` field type keeps the following elements:
-
-* br
-* p
-* big
-* em
-* small
-* strong
-* sub
-* sup
-* ins
-* del
-* code
-* kbd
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +44,93 @@ An item refers to a single item of data scraped from the target website. A commo
...
+        </tr>
+        <tr>
+            <td>date</td>
+            <td>A date value parsed by [dateparser](https://github.com/scrapinghub/dateparser). This won't work if the annotated element includes non-date text, in which case you should use partial annotations.</td>
+        </tr>
+    </tbody>
+</table>
+
+The ``safe html`` field type keeps the following elements:
+
```
redapple: The list of tags takes up a lot of space.
Could they be grouped somehow? or inlined?

redapple: Ok, fixed in new commit, while I was commenting :)

PR #206, opened 2015-06-05
Develop

YouWoTMA:
Improvements to the toolbox:
* Remember the pinned state of the toolbox using localStorage.
* Start with the toolbox open (if it's pinned, it doesn't waste time with unnecessary animation, if unpinned if gives the user visual feedback of where are the controls).
* Hide thumbtack if the toolbox is fixed, since it doesn't perform any action.


ruairif:
Thanks. I was just about to create a ticket for this :smile:

PR #219, opened 2015-06-30
Fix conflicts resolution and add support for merging lists

ruairif:


slyd/app/mixins/conflict-mixin.js
```diff
@@ -0,0 +1,131 @@
...
+        this._updatePath(path);
+    },
+
+    _updatePath: function(path) {
+        try {
+            this.sendAction('conflictOptionUpdated', path,
+                            this.get('conflictedKeyPaths.'+path+'.accepted'),
+                            this.get('conflictedKeyPaths.'+path+'.rejected'));
+        } catch (e) {
+            if (!/^TypeError/.test(e.toString())) {
```
michalmo: why not use instanceof?

slyd/slyd/gitstorage/jsondiff.py
```diff
@@ -1,4 +1,127 @@
...
+        return mine
+    if other == base != mine:
+        return mine
+    if mine == base != other:
+        return other
+    result = []
+    last_conflict = False
+    for i, (m, o, b) in enumerate(izip_longest(mine, other, base,
+                                               fillvalue=_BLANK)):
+        if m == o == b or m == o:
```
michalmo: same as `if m == o:` so the elif is the only alternative, and the else is never reached

slyd/slyd/gitstorage/jsondiff.py
```diff
@@ -1,4 +1,127 @@
...
+                result.append(c)
+            last_conflict = True
+            continue
+        else:
+            result.append(b)
+        last_conflict = False
+    for i, r in enumerate(result[:]):
+        if isinstance(r, Conflict):
+            c = r.resolve_conflict()
+            result = result[:i] + c
```
michalmo: if `len(c) > 1` then value of `i` will be off

slyd/slyd/gitstorage/jsondiff.py
```diff
@@ -1,4 +1,127 @@
...
+        return 'Conflict{}'.format(str((self.mine, self.other, self.base)))
+
+    def __repr__(self):
+        return str(self)
+
+
+def merge_lists(base, mine, other):
+    if mine == other:
+        return mine
+    if other == base != mine:
```
michalmo: `!= mine` and `!= other` below are redundant

slyd/slyd/gitstorage/projects.py
```diff
@@ -138,7 +138,8 @@ def edit_project(self, name, revision):
     @run_in_thread
     def publish_project(self, name, force):
         repoman = self._open_repo(name)
-        if repoman.publish_branch(self._get_branch(repoman), force):
+        if (repoman.publish_branch(self._get_branch(repoman),
+                                   force=force) is True):
```
michalmo: In the past I've had cases where True was not True, when the returned True was from a different module than the tested True. Python only guarantees that None is a singleton so this should be `== True`

slyd/slyd/gitstorage/projects.py
```diff
@@ -155,9 +156,8 @@ def project_revisions(self, name):
     @run_in_thread
     def conflicted_files(self, name):
         repoman = self._open_repo(name)
-        return json.dumps(
-            repoman.get_branch_conflicted_files(
-                self._get_branch(repoman, read_only=True)))
+        branch = self._get_branch(repoman, read_only=True)
+        return json.dumps(repoman.publish_branch(branch, dry_run=True))
```
michalmo: publish_branch can return True

slyd/slyd/gitstorage/repoman.py
```diff
@@ -416,16 +421,25 @@ def load_json(path, branch):
                     if my_changes.type == CHANGE_RENAME:
                         jsons[1] = load_json(my_changes.new.path, mine)
                         path = my_changes.new.path
-                    merged_json, merge_conflict = merge_jsons(*jsons)
-                    had_conflict = had_conflict or merge_conflict
+                    if take_mine:
+                        merged_json = jsons[1] or jsons[2] or jsons[0]
```
michalmo: these could use better names so that the order used is more obvious

PR #223, opened 2015-07-02
Css selectors backend changes

Youwotma:


slybot/slybot/plugins/selectors/__init__.py
```diff
@@ -0,0 +1,28 @@
+
+class Selectors(object):
+    def setup_bot(self, settings, spec, items, extractors):
+        self.selectors = {} # { item_type: { field_name: {..} }
+
+        for template in spec['templates']:
+            selectors = self.selectors.setdefault(template['scrapes'], {})
+            selectors.update(template.get('selectors', {}))
```
ruairif: This only allows a single css/xpath expression per field name per item type, right?

slybot/slybot/plugins/selectors/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+        item_type = item.get('_type', '')
+        selectors = self.selectors.get(item_type)
+        if not selectors:
+            return
+
+        for field, selector_data in selectors.iteritems():
+            selector = selector_data['selector']
+            selector_type = selector_data['type']
+            if selector_type == 'css':
+                item[field] = response.css(selector).xpath('./text()').extract()
```
ruairif: Should we concatenate the extracted data to the field if there is already data there?

slybot/slybot/plugins/selectors/__init__.py
Youwotma: Yes, selectors could be overwritten if there are set in more than one template that extracts the same item type.

Still, I think several selectors for one field is overkill, since CSS and xpath already have a way to do a union of several selectors (`//summary | //h1` and `summary, h1`). What do you think would be the best solution?
- We could put the item definitions in the item instead and all templates that extract the same item use that selectors
- Keep the selectors in the template definition and run only that selectors only when that template is the one that matches
- Keep it like it is now (merge the selectors of every template that extracts the same item and run all of them when any template extracts an item)

slybot/slybot/plugins/selectors/__init__.py
Youwotma: Good idea

slybot/slybot/plugins/selectors/__init__.py
ruairif: Using the selectors that match the current template is probably the safest but it needs a fallback if the annotations-plugin is disabled.

The fallback could be to try all templates and set the template used as the one that extracts the most data which is what the annotations-plugin does.

PR #224, opened 2015-07-06
Field name validation refactor

Youwotma:
Refactor validations:
- Show an error message when adding a duplicated field
- Show an error message when adding a field that doesn't start with a letter (to prevent overwriting _item and _template)
- Show an error when changing an item name if there is already a item with that name.
- Validate field names when creating them in the annotation editor too.

Also on this pull request:
- Added one test
- Fixed a couple sentry issues

### DIFF
--- a/slybot/docs/project.rst
+++ b/slybot/docs/project.rst
@@ -82,7 +82,7 @@ name : string
 
 version : string
   Version number of the format.
-  
+
 comment : string : optional
   A comment provided by the user.
 
@@ -112,6 +112,10 @@ system, i.e: person, job, category, etc.::
     },
 
 Attributes:
+
+displayName: string : optional
+  User-friendly name of the item. If not specified, the item id will be used
+  instead.
 
 fields : mapping
   This is a mapping of the field names to the field objects representing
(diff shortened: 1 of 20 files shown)


slyd/app/components/edit-item.js
```diff
@@ -1,41 +1,50 @@
 import Ember from 'ember';
 import NotificationHandler from '../mixins/notification-handler';
 
+export function validateFieldName(name, fields) {
+    // Ensuring that field names start with a letter prevents overwriting
+    // _item, _template and any future "protected" property we might
+    // add to extracted items.
+    if (!/^[a-z]/i.test(name)) {
```
ruairif: Maybe just disallow names beginning with `_`? A user may want to user a number as the first character.

slybot/docs/project.rst
```diff
@@ -113,6 +113,10 @@ system, i.e: person, job, category, etc.::
 
 Attributes:
 
+displayName: string : optional
```
ruairif: Use `display_name` so that it matches the other property names

PR #232, opened 2015-07-15
Sentry errors

Youwotma:


slyd/slyd/splash/proxy.py
```diff
@@ -20,31 +22,55 @@ def render_GET(self, request):
...
 
-    def end_response(self, request, original_url, tabid, reply):
+    def _requestError(self, err, request):
+        if not err.check(CancelledError):
+            request.setResponseCode(500)
+            request.write('Error fetching the content')
+            request.finish()
+
+    def _requestDisconnect(self, err, deferred=None, connection_status=None):
+        print "_requestDisconnect"
```
ruairif: Is this for debug or to keep track of when this is called?

Youwotma: For debug, forgot to remove it. Will update the PR.

PR #238, opened 2015-07-24
Fix several XSS vulnerabilities

Youwotma:
Can be tested in this page for event and href vulnerabilities:
http://dev.scrapinghub.com/~david/xsstest.html
And for meta redirect:
http://dev.scrapinghub.com/~david/xsstest2.html

Some worked in interactive (browsing) mode, some in annotation mode, some in both.

### DIFF
--- a/slyd/slyd/html.py
+++ b/slyd/slyd/html.py
@@ -22,19 +22,22 @@ from .utils import serialize_tag, add_tagids
 URI_ATTRIBUTES = ("action", "background", "cite", "classid", "codebase",
                   "data", "href", "longdesc", "profile", "src", "usemap")
 
-AS_SCRIPT_REGION_BEGIN = "<!-- begin region added by slyd-->"
-AS_SCRIPT_REGION_END = "<!-- end region added by slyd-->"
+# https://html.spec.whatwg.org/multipage/syntax.html#character-references
+# http://stackoverflow.com/questions/18689230/why-do-html-entity-names-with-dec-255-not-require-semicolon
+_ENTITY_RE = re.compile("&#(\d+[^\d]|x[a-f\d]+[^a-f\d])", re.I)
 
-_AS_COMMENT_BEGIN = "<!-- begin_ascomment:"
-_AS_COMMENT_END = ":end_ascomment -->"
-_ENTITY_RE = re.compile("&#(\d+);")
-
+def _replace_entity(match):
+    entity = match.group(1)
+    if entity[0].lower() == 'x':
+        return unichr(int(entity[1:-1], 16))
+    else:
+        return unichr(int(entity[0:-1], 10))
 
 def _deentitize_unicode(mystr):
     """replaces all entities in the form &#\d+; by its
     unicode equivalent.
     """
-    return _ENTITY_RE.sub(lambda m: unichr(int(m.groups()[0])), mystr)
+    return _ENTITY_RE.sub(_replace_entity, mystr)
 
 
 def html4annotation(htmlpage, baseurl=None, proxy_resources=None):
@@ -82,7 +85,7 @@ def descriptify(doc, base=None, proxy=None):
                         element.attributes[key] = process_css(val, -1, base)
                     # Rewrite javascript URIs
                     elif key in URI_ATTRIBUTES and val is not None:
-                            if "javascript:" in _deentitize_unicode(val):
+                            if "javascript:" in _deentitize_unicode(val).lower():
                                 element.attributes[key] = "about:blank"
                             elif base and proxy and not (element.tag == "a" and key == 'href'):
                                 element.attributes[key] = wrap_url(val, -1,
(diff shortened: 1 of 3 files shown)


slyd/slyd/html.py
```diff
@@ -82,13 +74,13 @@ def descriptify(doc, base=None, proxy=None):
...
                     # Empty intrinsic events
-                    if key in INTRINSIC_EVENT_ATTRIBUTES:
+                    if key.startswith('on') or key == "http-equiv":
                         element.attributes[key] = ""
                     elif base and proxy and key == "style" and val is not None:
                         element.attributes[key] = process_css(val, -1, base)
                     # Rewrite javascript URIs
                     elif key in URI_ATTRIBUTES and val is not None:
-                            if "javascript:" in _deentitize_unicode(val):
+                            if "javascript:" in _deentitize_unicode(val).lower():
```
ruairif: The improvements for catching xss with `javascript:` are good but they miss out on some cases listed [here](https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet).

I've put together another solution for handling this. It involves using an unescape function derived from the one in the html module of python3.4 when using python2. It also limits the allowed characters so that we can avoid issues with newlines, vertical tabs, horizontal tabs etc:
```python
import re
_ALLOWED_CHARS_RE = re.compile('[^a-z:0-9;,"\']', re.IGNORECASE).sub
protect_xss = lambda s: _ALLOWED_CHARS_RE('', s)
_ENTITY_RE = re.compile('&(#[0-9]+;?|#[xX][0-9a-fA-F]+;?|[^\\t\\n\\f <&#;]{1,32};?)', re.IGNORECASE)


try:
    from html import unescape

    _deentitize_unicode = unescape
except ImportError:
    def _replace_entity(match):
        entity = match.group(1)
        if entity[0] == '#':
            if entity[1] in 'xX':
                num = int(entity[2:].rstrip(';'), 16)
            else:
                num = int(entity[1:].rstrip(';'))
            if 0xD800 <= num <= 0xDFFF or num > 0x10FFFF:
                return '\uFFFD'
            return unichr(num)
        else:
            return '&' + match

    def _deentitize_unicode(mystr):
        if '&' not in s:
            return s
        return _ENTITY_RE.sub(_replace_entity, mystr)
```

Here are the test cases:
```python
strs = [
    "javascript:alert('XSS');",
    "javascript:alert('XSS')",
    'JaVaScRiPt:alert("XsS");',
    '&#0000106&#0000097&#0000118&#0000097&#0000115&#0000099&#0000114&#0000105&#0000112&#0000116&#0000058&#0000097&#0000108&#0000101&#0000114&#0000116&#0000040&#0000039&#0000088&#0000083&#0000083&#0000039&#0000041',
    '&#x6A&#x61&#x76&#x61&#x73&#x63&#x72&#x69&#x70&#x74&#x3A&#x61&#x6C&#x65&#x72&#x74&#x28&#x27&#x58&#x53&#x53&#x27&#x29',
    "jav        ascript:alert('XSS');",
    "jav&#x09;ascript:alert('XSS');",
    "jav&#x0A;ascript:alert('XSS');",
    "jav&#x0D;ascript:alert('XSS');",
    'java\0script:alert(\"XSS\")>";',
    " &#14;  javascript:alert('XSS');",
]

["javascript:" in protect_xss(_deentitize_unicode(s)).lower() for s in strs]
```
What do you think?

Youwotma: Sounds good, I'll add the test cases to the tests and update the PR.
Is protect_xss function necessary? It looks over-protective to me since replaces characters like /, &, ? and % that are normal to find in an url.

ruairif: It could probably be expanded to allow more characters or we could just remove control characters and strip the string instead.

slyd/slyd/html.py
```diff
@@ -15,34 +15,30 @@
...
+try:
+    from html import unescape
+except ImportError:
+    # https://html.spec.whatwg.org/multipage/syntax.html#character-references
+    # http://stackoverflow.com/questions/18689230/why-do-html-entity-names-with-dec-255-not-require-semicolon
+    _ENTITY_RE = re.compile("&#(\d+|x[a-f\d]+);?", re.I)
+    def _replace_entity(match):
+        entity = match.group(1)
+        if entity[0].lower() == 'x':
+            return unichr(int(entity[1:], 16))
```
ruairif: Should we use `six.unichr`?

slyd/slyd/html.py
```diff
@@ -15,34 +16,30 @@
...
-_AS_COMMENT_END = ":end_ascomment -->"
-_ENTITY_RE = re.compile("&#(\d+);")
-
-
-def _deentitize_unicode(mystr):
-    """replaces all entities in the form &#\d+; by its
-    unicode equivalent.
-    """
-    return _ENTITY_RE.sub(lambda m: unichr(int(m.groups()[0])), mystr)
+_ALLOWED_CHARS_RE = re.compile('[^a-z:0-9;,"\'\\/?&%#]', re.IGNORECASE)
```
ruairif: What about allowing these characters for the URL?

    -._~[]@!$()*+=

PR #247, opened 2015-08-06
Added docs on annotating elements contained within <head>

rdowinton:
Some users have recently been asking about how to annotate elements contained within the page's head element, so I thought it would be useful to document.


ruairif:
This has been changed with the latest release of Portia. Updated docs are available in the README

PR #249, opened 2015-08-08
updated docker run command

e6:


Youwotma:
Hey @e6 thanks for the PR! we are in the middle of deploying a new version of Portia, that's why it's taking so long to merge, sorry! :)

ruairif:
Thank you for the correction, sorry it took so long to merge.

PR #250, opened 2015-08-20
Connect new UI browser view port to slyd splash browser

Youwotma:
Allows loading the page, needs to forward events and update browser history.


slyd/slyd/resource.py
```diff
@@ -65,5 +65,6 @@ class SlydJsonObjectResource(SlydJsonResource):
 
     def render(self, request):
         resp = SlydJsonResource.render(self, request)
-        if resp is not None:
-            return json.dumps(resp)
+        if type(resp) == dict:
```
ruairif: Why not use isinstance?

Youwotma: will do!

PR #257, opened 2015-08-25
Fix broken pages with malformed tag names, toggleCSS and duplicate template names

Youwotma:
- Fix broken pages with malformed tag names like `<imgsrc="image.png"/>`. Also indirectly fixes some of the "No node with id \d+" errors.
- Make sure template names are unique
- Reset the state of cssEnabled when changing modes, read the state from documentView


Youwotma:
Also two more fixes

PR #265, opened 2015-09-02
Fix "re.match object is not serializable" and correctly read the...

Youwotma:
... _uses_js property


Youwotma:
Closing because selected wrong branch to merge into

PR #267, opened 2015-09-03
Couple fixes

Youwotma:


slyd/slyd/html.py
```diff
@@ -15,7 +15,7 @@
 from .splash.css_utils import process_css, wrap_url
 from .utils import serialize_tag, add_tagids
 
-URI_ATTRIBUTES = ("action", "background", "cite", "classid", "codebase",
+URI_ATTRIBUTES = ("action", "background", "cite", "classid", "codebase", "action",
```
ruairif: Why is action in there twice?

Youwotma: My mistake, didn't see it

PR #235, opened 2015-07-21
Record page actions and apply them when crawling

Youwotma:
Looking for feedback about this in general and specifically: For replying actions some code needs to be loaded into the page, at the moment this code is sent with every request to splash, is there a better way to get around this?

PR #280, opened 2015-09-15
Add ES5-shim and localStorage shim to Slybot's Splash requests.

Youwotma:
The combined file it's compiled with a script and checked in in a
repository. It would be better if it wasn't checked in and Travis built
it when deploying Slybot to pypy instead.

Note: We can later use this same file to add two scripts required by 
page-actions, so should be merged before #235 and then I'll update 
the other PR.


Youwotma:
Cancelled because I sent it to the wrong branch

PR #292, opened 2015-10-01
Strip null characters before decoding JSON

Youwotma:


slyd/slyd/splash/ferry.py
```diff
@@ -136,10 +136,13 @@ def log(self, s):
     @pyqtSlot('QString')
     def sendMessage(self, message):
         message = text(message)
+        message = message.strip('\x00') # Allocation bug somewhere leaves null characters at the end.
         try:
             command, data = json.loads(message)
-        except ValueError:  # XXX: Possibly null terminated string
-            command, data = json.loads(message[:-1])
+        except ValueError as e:
+            return log.err(ValueError(
```
michalmo: id prefer the traceback to go to sentry, it would allow to see where this is coming from.
just `log.err()` here will create a  twisted failure with the current exc_info()

Youwotma: This is called from javascript, so the traceback is just a couple of pyqt
frames and nothing from our app, not very useful.

PR #330, opened 2015-12-28
Splash Qt5 Support

Youwotma:


Vagrantfile
```diff
@@ -1,14 +1,13 @@
...
 Vagrant.configure("2") do |config|
- 	config.vm.box_url = 'http://files.vagrantup.com/precise32.box'
-	config.vm.box = "precise32"
+	config.vm.box = "ubuntu/trusty64"
 	config.vm.host_name = "portia"
-	config.vm.provision :shell, :path => 'provision.sh'
+	config.vm.provision :shell, :path => 'provision.sh', :args => ["install_deps", "install_splash", "install_python_deps", "configure_nginx", "configure_initctl"]
 	config.vm.network "private_network", ip: "33.33.33.10"
-	config.vm.network "forwarded_port", guest: 8000, host: 8000
+	config.vm.network "forwarded_port", guest: 9001, host: 9001
```
ruairif: Is there any reason to change this? People who currently use this already have their links set for port 8000 and might be confused when it stops working.

provision.sh
```diff
@@ -1,31 +1,138 @@
...
+    pip install http://cdn.mysql.com/Downloads/Connector-Python/mysql-connector-python-1.2.3.zip#md5=6d42998cfec6e85b902d4ffa5a35ce86
+    pip install -r "$APP_ROOT/slyd/requirements.txt"
+    pip install -r "$APP_ROOT/slybot/requirements.txt"
+    pip install -e "$APP_ROOT/slyd"
+    pip install -e "$APP_ROOT/slybot"
+}
+
+install_splash(){
+    cd /tmp
+    curl -L -o splash.tar.gz 'https://github.com/scrapinghub/splash/archive/master.tar.gz'
```
ruairif: Will this be tied to a release branch at some point?

Youwotma: Yes, once splash 2.0 is out. We should think about how to deploy this, if we want to release for the new UI only or make an old UI release with Qt5. Either way should cause no problem, just merging this pull request into the new UI branch should work since this doesn't change anything in the UI.

Vagrantfile
Youwotma: Ok, just changed it because it was conflicting with something on my system

Youwotma: I changed the vagrant image to use nginx too like the docker image, this was so it's easier to maintain both different images (being more similar and provisioning from a single script).

I can change it back if there are reasons to change it back.

Vagrantfile
```diff
@@ -1,14 +1,13 @@
...
 	config.vm.provider "virtualbox" do |v|
-		v.memory = 512
+		v.memory = 2048
```
Youwotma: Also this is because it has to build pyqt from source, and it needs the memory for linking. When the appropriate versions of pyqt are in the repositories we can change it back

slyd.conf
```diff
@@ -3,5 +3,7 @@ start on vagrant-mounted
 stop on runlevel [!2345]
 
 chdir /vagrant/slyd
-exec bin/slyd -p 8000 --root ./dist
+script
+    PYTHONPATH=/vagrant/slyd:/vagrant/slybot bin/slyd -p 9002 --root ./dist
```
vgvassilev: I needed to change this here to 9001, otherwise the service wouldn't show up on localhost:9001

Dockerfile
```diff
@@ -1,33 +1,19 @@
...
-
-RUN ln -sf /app/nginx/nginx.conf /etc/nginx/nginx.conf
-
-RUN pip install -e /app/slybot /app/slyd
+EXPOSE 9001
 
 WORKDIR /app/slyd
+
 # TODO(dangra): fix handling of nginx service, it won't be restarted in case if crashed.
 CMD service nginx start; bin/slyd -p 9002 -r /app/slyd/dist
```
vgvassilev: I assume this is the case here too. The service wouldn't show up on localhost:9001

slyd.conf
Youwotma: Hi! That was because nginx wasn't starting properly, should be fixed now

PR #340, opened 2016-01-18
Add support for atom, rss and xml sitemaps

ruairif:


slybot/slybot/plugins/scrapely_annotations/annotations.py
```diff
@@ -211,8 +214,16 @@ def _request_to_follow_from_region(self, htmlregion):
             if request is not None:
                 yield request
 
-    def handle_rss(self, response, seen):
-        for link in self.rss_link_extractor.links_to_follow(response):
```
Youwotma: Is `self.rss_link_extractor` still used anywhere? If it isn't remove it from `setup_bot()` as well

slybot/slybot/spider.py
```diff
@@ -197,14 +192,15 @@ def parse(self, response):
...
-        elif "application/rss+xml" in content_type:
-            return self.handle_rss(response)
-        else:
-            self.logger.debug(
-                "Ignoring page with content-type=%r: %s" % (content_type,
-                                                            response.url)
-            )
-            return []
+        if (isinstance(response, XmlResponse) or
+                response.url.endswith(('xml', 'xml.gz'))):
```
Youwotma: A bit unsafe, a lot of html pages about xml will end in xml. Can we use `.endswith(('.xml', '.xml.gz'))` (notice the extra dots) so it's less likely?

PR #350, opened 2016-01-22
Update installation.rst

flip111:
Somewhere in 2015 a requirement for a certain https://python-pillow.github.io/ version made it so that the development headers for jpeg are also needed. This was tested on ubuntu 14.04 raspberry pi 2 model B. Error log will be up for a month http://pastebin.com/R1S17NCk


ruairif:
Should now be installed by splash

PR #367, opened 2016-02-04
Add and remove repeated container correctly.

ruairif:
Use body tag as default container selector.
Add tracebacks for websocket errors.


portiaui/app/services/dispatcher.js
```diff
@@ -360,7 +360,9 @@ export default Ember.Service.extend({
         }
         const parent = annotation.get('parent.itemAnnotation');
         annotation.destroyRecord().then(() =>
-            this.updateContainers(parent)
+            Ember.run.next(this, () =>
+                this.updateContainers(parent)
```
michalmo: updateContainers() already call Ember.run.next inside it's body

PR #370, opened 2016-02-09
Add backwards compatability for spiders

ruairif:
Add selectors to existing annotations.
Add pre and post text to generated annotations.
Use `name` field rather than display name for item and field names
When loading template add annotation text extractor to annotations with pre and post text.
266 Add support for nested items.
267 Add annotation from dropdown without error.
284 Add variant extractors.
292 Add extractors backend.
293 Auto open link rules when user changes to patterns or all.
Add per annotation extractors rather than per field.
Change annotation format to allow extracting a single attribute to many fields


slybot/slybot/extractors.py
```diff
@@ -61,3 +71,14 @@ def apply_extractors(descriptor, template_extractors, extractors):
         if equeue:
             equeue.insert(0, descriptor.attribute_map[field_name].extractor)
             descriptor.attribute_map[field_name].extractor = PipelineExtractor(*equeue)
+
+def add_extractors_to_descriptors(descriptors, extractors):
+    new_extractors = {}
+    for id, data in extractors.items():
```
plafl: `id` is a built-in function, maybe another identifier would be better

slybot/slybot/extractors.py
```diff
@@ -24,6 +24,15 @@ def _extractor(txt):
     _extractor.__name__ = "Regex: %s" % pattern.encode("utf-8")
     return _extractor
 
+def create_type_extractor(type):
+    fieldtypes = FieldTypeManager()
+    extractor = types.type_processor_class(type)()
+    def _extractor(txt):
```
plafl: `_extractor` accepts `txt` argument, but is not used

slybot/slybot/plugins/scrapely_annotations/builder.py
```diff
@@ -54,11 +55,17 @@ def _clean_annotation_data(data):
         if ann.get('item_container'):
             ann['annotations'] = {'#portia-content': '#dummy'}
             ann['text-content'] = '#portia-content'
-            result.append(ann)
+        elif 'data' in ann:
+            modified_annotaions = {}
```
plafl: typo: `modified_annotaions` should be `modified_annotations`

slybot/slybot/plugins/scrapely_annotations/extraction.py
```diff
@@ -178,8 +229,9 @@ def _get_container_data(extractors):
         non_container_annos = []
         for con_id, annotation in groupby(extractors, container_id):
             annotation = list(annotation)
-            if (con_id and str(con_id).endswith('#parent')
-                    and con_id not in containers):
+            # XXX: Handle repeated container deletion but not parent container
```
plafl: `XXX`? Maybe TODO? :)

slybot/slybot/plugins/scrapely_annotations/extraction.py
```diff
@@ -96,6 +108,43 @@ def descriptor(self, descriptor_name=None):
...
+        """
+        if ignored_regions is None:
+            ignored_regions = []
+        extractors = sorted(self.extractors + ignored_regions,
+                            key=lambda x: labelled_element(x).start_index)
+        _, _, attributes = self._doextract(page, extractors, start_index,
+                                           end_index, **kwargs)
+        # collect variant data, maintaining the order of variants
+        items = []
+        for k, v in attributes:
```
plafl: The `k, v` identifiers give the impression that `attributes` is a dict, in which case the correct way of iteration would be:
```
for k, v in attributes.iteritems()
```

If that is not the case and `attributes`is a list or other iterable then the iteration is not needed and should just simply do `items = list(attributes)`

portiaui/app/components/list-item-link-crawling.js
```diff
@@ -33,7 +33,11 @@ export default Ember.Component.extend({
 
     actions: {
         saveSpider() {
-            this.get('spider').save();
+            this.get('spider').save().then(() => {
+                if (this.get('linksToFollow') !== 'none') {
```
michalmo: shouldn't this only open when `=== 'patterns'` ?

portiaui/app/routes/projects/project/spider.js
```diff
@@ -54,5 +54,9 @@ export default Ember.Route.extend({
             this.transitionTo('projects.project',
                 this.modelFor('projects.project'));
         }
+
+        openLinkOptions() {
```
michalmo: can we move this to link-item-list-crawling?
you can use `Ember.inject.service('-routing')` which has the `transitionTo` method there, instead of passing the action around.

portiaui/app/components/spider-structure-listing.js
```diff
@@ -44,6 +44,14 @@ export default Ember.Component.extend({
 
         saveSample(sample) {
             sample.save();
+        },
+
+        openLinkOptions() {
+            this.sendAction('openLinkOptions');
+        },
+
+        closeLinkOptions() {
```
michalmo: unused

portiaui/app/components/list-item-link-crawling.js
ruairif: When they set it to all they can set the nofollow rule too. Should I change to to just open for patterns?

slyd/slyd/resources/annotations.py
```diff
@@ -44,61 +45,58 @@ def update_annotation(manager, spider_id, sample_id, annotation_id,
...
+                                         anno_id)
+    annotation_data = annotation['data'][_id]
+    field_id = annotation_data['field']
+    if relationships.get('field_id'):
+        field_id = relationships['field_id']
+    annotation_data['field'] = field_id
+    attribute = annotation_data['attribute']
+    if data.get('attribute'):
+        attribute = data['attribute']
+    annotation_data['attribute'] = attribute
```
michalmo: if attribute is not present on data, you just read it and write it back in

slyd/slyd/resources/annotations.py
```diff
@@ -44,61 +45,58 @@ def update_annotation(manager, spider_id, sample_id, annotation_id,
...
+    if data.get('required'):
+        annotation_data['required'] = True
+    else:
+        annotation_data['required'] = False
```
michalmo: `annotation_data['required'] = data.get('required', False)`

portiaui/app/components/list-item-link-crawling.js
michalmo: my guess would be the user only expects to see the options when they want to set patterns.
in the other case it may be annoying espacialy if they switch to 'none' and then back to 'all'.
lets go with just showing it for patterns for now, and wait for feedback

slyd/slyd/resources/extractors.py
```diff
@@ -0,0 +1,79 @@
...
+def get_extractor(manager, extractor_id, attributes=None):
+    extractors = _read_extractors(manager)
+    extractor = extractors[extractor_id]
+    return ExtractorSchema(context=ctx(manager)).dump(extractor).data
+
+
+def create_extractor(manager, attributes):
+    extractors = _read_extractors(manager)
+    _id = gen_id(disallow=set(extractors.keys()))
+    attrs = attributes.get('data', {}).get('attributes', {})
```
michalmo: should this be setdefault? the attrs dict will be modified in-place if it exists, but a new one will be used and discarded if not

slyd/slyd/resources/extractors.py
```diff
@@ -0,0 +1,79 @@
...
+    extractor = _build_extractor(manager, attrs, _id)
+    extractors[_id] = extractor
+    manager.savejson(extractors, ['extractors'])
+    return ExtractorSchema(context=ctx(manager)).dump(extractors[_id]).data
+
+
+def update_extractor(manager, extractor_id, attributes):
+    extractors = _read_extractors(manager)
+    extractor = extractors[extractor_id]
+    attrs = attributes.get('data', {}).get('attributes', {})
```
michalmo: as above

slyd/slyd/utils/migration.py
```diff
@@ -0,0 +1,408 @@
...
+    a_parents = list(a.iterancestors())
+    b_parents = list(b.iterancestors())
+    a_parents_set = set(a_parents)
+    b_parents_set = set(b_parents)
+    if a == b:
+        return a
+    if b in a_parents_set:
+        return b
+    if a in b_parents_set:
+        return b
```
michalmo: return a?

slyd/slyd/resources/extractors.py
ruairif: attrs is loaded from attributes. Even with setdefault it would be necessary to have a second setdefault to be sure it wouldn't throw a key error.
We could do this instead too:
```python
try:
    attrs = attributes['data']['attributes']
except KeyError:
    attrs = {}

slyd/slyd/resources/annotations.py
ruairif: An attribute should always exist otherwise the annotation is useless since it won't extract any data.

PR #397, opened 2016-02-29
Nui fixes

Youwotma:


portiaui/app/components/inspector-panel.js
```diff
@@ -45,6 +45,26 @@ export function getAttributeList(element) {
     return attributeList;
 }
 
+export function getDefaultAttribute(element) {
+    // TODO: Remove attributes that have been already annotated
```
ruairif: Attributes can be annotated multiple times. Would you be removing them just to make selecting the default attribute more useful?

portiaui/app/services/dispatcher.js
```diff
@@ -205,14 +205,11 @@ export default Ember.Service.extend({
         });
         if (element) {
             annotation.addElement(element);
+            attribute = attribute || getDefaultAttribute(element);
```
michalmo: can you make it set the annotation type to url and image for these cases too?

portiaui/app/components/inspector-panel.js
Youwotma: True, not sure what the most common use case would be, if annotating the same attribute twice, or annotating different attributes. There are few cases where you want different attributes (href and content, src and title) but also few use cases for annotating the same attribute twicd.

At the moment when an element is already annotated the default action with the magic tool enabled will be to select the existing annotation anyway

michalmo:
needs discussion on spider renaming

PR #408, opened 2016-03-03
added support for error logging

michalmo:


portiaui/app/instance-initializers/error-handler.js
```diff
@@ -32,20 +45,13 @@ export function initialize(applicationInstance) {
...
-                    'Details have been logged to the console.',
+                message: 'An unexpected error has occurred. ' + instructions,
                 type: 'danger'
             });
         }
     }
 
     Ember.onerror = notifyError;
-
-    applicationInstance.ApplicationRoute = Ember.Route.extend({
```
michalmo: removed since this is causing double logging now. probably ember 2 calls Ember.onerror.

PR #415, opened 2016-03-10
Slybot 0.13.0b11

ruairif:
Add start urls generation
Fix per annotation requireds
Fix illegal characters in urls
Change speed test iterations using `SLYBOT_SPEED_TEST_ITERATIONS` env variable
Refactor spider init
Improve tests


plafl:
Looks good to me

PR #436, opened 2016-03-25
Add git status endpoint for projects

ruairif:
`GET portia/api/projects/<project_id>/status`


slyd/slyd/gitstorage/projects.py
```diff
@@ -172,12 +172,15 @@ def conflicted_files(self, name):
 
     @run_in_thread
     def changed_files(self, name):
-        return self._changed_files(name)
+        return json.dumps([fname for _, fname, _ in self._changed_files(name)])
```
michalmo: should be `fname of oldname`

PR #448, opened 2016-04-04
Add UI for auto pagination

ruairif:
Add auto option to link crawling rules
Fix case where repeated not set correctly for item_annotation

### DIFF
--- a/portiaui/app/models/item-annotation.js
+++ b/portiaui/app/models/item-annotation.js
@@ -8,11 +8,6 @@ export default Annotation.extend({
         async: true,
         inverse: 'itemAnnotation'
     }),
-    repeated: DS.attr('boolean', {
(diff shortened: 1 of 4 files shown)


portiaui/app/models/item-annotation.js
```diff
@@ -8,11 +8,7 @@ export default Annotation.extend({
         async: true,
         inverse: 'itemAnnotation'
     }),
-    repeated: DS.attr('boolean', {
-        defaultValue() {
-            return false;
-        }
-    }),
+    repeated: Ember.computed.bool('repeatedAcceptSelectors'),
```
michalmo: shouldn't this be `Ember.computed.bool('repeatedAcceptSelectors.length')` ?

PR #459, opened 2016-04-18
vagrant compatible

01-:


ruairif:
Sorry but these changes are specific to your system.
The IP is your own local ip and the DEFAULT_PORTIA_ROOT is where the css and js assets are stored.

PR #470, opened 2016-05-04
Change how portia writes to DB

ruairif:
Update all sample sub assets with sample
Set selection_mode when updating annotation
Update to latest dulwich
Add custom dulwich mysql backend to portia
Change websocket to have managers built in just like api
Run each request in its own thread if using mysql backend
Buffer writes to db

### DIFF
--- a/portiaui/app/serializers/sample.js
+++ b/portiaui/app/serializers/sample.js
@@ -16,7 +16,7 @@ export default DS.JSONAPISerializer.extend({
     },
 
     _addAnnotations(item, includes) {
-        let data =item.belongsTo('itemAnnotation').serialize().data;
+        let data = item.belongsTo('itemAnnotation').serialize().data;
         data.id = item.id;
         includes.push(data);
         for (let annotation of item.hasMany('annotations')) {
--- a/portiaui/app/serializers/sample.js
+++ b/portiaui/app/serializers/sample.js
@@ -1,0 +1,32 @@
+import DS from 'ember-data';
+
+export default DS.JSONAPISerializer.extend({
+    serialize(snapshot, options) {
+        let data = this._super(...arguments);
+        data.includes = this.serializeIncludes(snapshot, options);
+        return data;
+    },
+
+    serializeIncludes(snapshot, options) {
+        let includes = [];
+        for (let item of snapshot.hasMany('items')) {
+            this._addAnnotations(item, includes);
+        }
+        return includes
+    },
+
+    _addAnnotations(item, includes) {
+        let data =item.belongsTo('itemAnnotation').serialize().data;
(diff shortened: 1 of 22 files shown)


portiaui/app/serializers/sample.js
```diff
@@ -0,0 +1,32 @@
...
+    serializeIncludes(snapshot, options) {
+        let includes = [];
+        for (let item of snapshot.hasMany('items')) {
+            this._addAnnotations(item, includes);
+        }
+        return includes
+    },
+
+    _addAnnotations(item, includes) {
+        let data =item.belongsTo('itemAnnotation').serialize().data;
```
michalmo: ` = `

slyd/slyd/api.py
```diff
@@ -77,7 +77,10 @@ def render(self, request):
                                   for k, v in parsed.named.items()}
                         parsed['attributes'] = load_attributes(request)
 
-                        api_response = callback(manager, **parsed)
+                        api_response = getattr(
```
michalmo: lets make run defined on the base manager classes as a noop and explicitly call it here

slyd/slyd/splash/ferry.py
```diff
@@ -209,22 +221,32 @@ def user(self):
...
     def onMessage(self, payload, isbinary):
+        pool = getattr(Repoman, 'pool', None)
         payload = payload.decode('utf-8')
         data = json.loads(payload)
+        project = data.get('project', data.get('_meta', {}).get('project'))
+        self.manager = self.spec_manager.project_spec(project, self.auth_info)
+        self.manager.pm = self.spec_manager.project_manager(self.auth_info)
+        if pool is None:
+            return wrap_message_callback(self, data)
+        return pool._runWithConnection(wrap_message_callback, self, data)
```
michalmo: wrap_message_callback takes a connection as the first argument
also runWithConnection should be used instead of _runWithConnection to not block the thread.

i propose that wrap_message_callback just return the response message, then you can change the above to
```python
params = []
if pool is not None:
    params.append(pool.runWithConnection)
params.extend([wrap_message_callback, connection, self, data])
deferred = twisted.internet.defer.maybeDeferred(*params)
deferred.addCallback(self.sendMessage)
```

the wrap_callback from gitstorage/projects could be reused here instead of wrap_message_callback

michalmo:
you also need to turn off autocommit and set the transaction isolation mode to REPEATABLE READ, and start a transaction when the connection is opened.

slyd/slyd/splash/ferry.py
michalmo: also the return value of onMessage is ignored by the base class

PR #477, opened 2016-05-17
PORTIA-470: add annotated_body to sample if not present

michalmo:


slybot/slybot/plugins/scrapely_annotations/annotations.py
```diff
@@ -83,15 +86,20 @@ def setup_bot(self, settings, spec, items, extractors):
...
         _links_item_descriptor = create_slybot_item_descriptor({'fields': {}})
         self._links_ibl_extractor = InstanceBasedLearningExtractor(
             [(t, _links_item_descriptor) for t in _links_pages]) \
             if _links_pages else None
 
         self.build_url_filter(spec)
 
+    def _get_annotated_template(self, template):
+        if template.get('version', '0.12.0') >= '0.13.0':
+            if 'annotated_body' not in template:
```
ruairif: It should build the sample even if annotated body is present as `annotated_body` is not updated in portia 2.0

PR #486, opened 2016-05-29
PORTIA-516: Template for browser selection

sagelliv:
@ruairif @michalmo 
Added a route `browsers` and template to display the browsers logos which allows the user to navigate to the download page for each browser.

browsers


portiaui/app/styles/templates/browsers.scss
```diff
@@ -0,0 +1,18 @@
+.browser-logos {
+    height: 120px;
+}
+
+.browser-logo {
+    height: 120px;
+    margin-right: 40px;
+    transition: 0.2s;
+    &:hover {
+        margin-top: -8px;
```
michalmo: can you animate this using translate instead of margin-top

portiaui/app/styles/templates/browsers.scss
```diff
@@ -0,0 +1,18 @@
+.browser-logos {
+    height: 120px;
+}
+
+.browser-logo {
+    height: 120px;
+    margin-right: 40px;
```
michalmo: can you add symmetric padding on both sides?

portiaui/app/styles/templates/browsers.scss
```diff
@@ -0,0 +1,18 @@
+.browser-logos {
+    height: 120px;
+}
+
+.browser-logo {
+    height: 120px;
```
michalmo: lets make these images size based on the container width so that they always fix next to each other on all screen sizes, and set this as a max-height

sagelliv:
@michalmo Fixed CSS to change sizes depending on the width of the browser container.

sagelliv:
Tested feature detection redirection on Opera 12 and it works. Made a more user-friendly message for users with unsupported features in their browsers. Left the browsers' size as 120x120 as this scales quite well in various sizes and they will just stack for quite small sizes.

portiaui/app/routes/index.js
```diff
@@ -1,7 +1,14 @@
 import Ember from 'ember';
+import hasBrowserFeatures from '../utils/browser-features';
+
+function identity(x) { return x; }
 
 export default Ember.Route.extend({
     redirect() {
-        this.transitionTo('projects');
+        hasBrowserFeatures().then((features) => {
+            let hasFeatures = features.every(identity);
```
michalmo: the documentation for redirect doesn't suggest that it follows the promise semantics of model/afterModel, so let's return hasBrowserFeatures from model() so that the route waits for it to resolve, and just transitionTo in redirect() based on the resolved value

sagelliv:
select-browsers
@michalmo

PR #488, opened 2016-06-03
Add optinonal page_clustering to slybot

plafl:
When PAGE_CLUSTERING setting is enabled (and extra `page_clustering` module has been installed), the most probable template according to the clustering algorithm will be suggested as the preferred template for item extraction.

### DIFF
--- a/slybot/slybot/plugins/scrapely_annotations/annotations.py
+++ b/slybot/slybot/plugins/scrapely_annotations/annotations.py
@@ -91,6 +91,16 @@ class Annotations(object):
             if _links_pages else None
 
         self.build_url_filter(spec)
+        # Clustering
+        self.template_names = [t.get('page_id') for t in spec['templates']]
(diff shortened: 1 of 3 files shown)


slybot/slybot/plugins/scrapely_annotations/annotations.py
```diff
@@ -91,6 +91,16 @@ def setup_bot(self, settings, spec, items, extractors):
...
 
         self.build_url_filter(spec)
+        # Clustering
+        self.template_names = [t.get('page_id') for t in spec['templates']]
+        if settings.get('PAGE_CLUSTERING'):
+            try:
+                import page_clustering
+                self.clustering = page_clustering.kmeans_from_samples(spec['templates'])
+            except ImportError:
+                self.clustering = None
```
ruairif: Can you add some logging here explaining that clustering wasn't and why

PR #492, opened 2016-06-05
Add lupa dependancy to slyd

G-Square:
added lupa in python req.
next pull will have scrapy.spider to scrapy.spiders fix.


ruairif:
It's only a warning and doesn't cause any problems. It can be added if lua scripting is needed for interacting with splash

PR #503, opened 2016-06-10
More fixes for switching schemas in a sample

michalmo:


slyd/slyd/gitstorage/projectspec.py
```diff
@@ -89,3 +89,8 @@ def commit_changes(self):
             return
         repo = self._open_repo()
         repo.save_files(self._changed_file_data, self._get_branch())
+
+    def rollback_changes(self):
+        if not self._changed_file_data:
```
ruairif: Why not just clear it?

PR #505, opened 2016-06-14
PORTIA-529: Generated Start URLs

sagelliv:
Feature: Users can now create generated start urls. A generated start url is a url which is made up by fragments. Some of these fragments can be lists or ranges which describe the generated start url as a set of urls, as opposed to fixed urls. The UI has been created according to [this moqup](https://app.moqups.com/scrapinghub/LgdhnSN7Tf/edit/page/aeb416392). Currently, the generated list of urls are generated with an inefficient but simple algorithm. In order for the user to be able to browse the generated list of urls, we should implement the algorithm using a generator in order to just generate the urls the user is seeing in the screen.
@michalmo @ruairif


portiaui/app/templates/components/start-url-options.hbs
```diff
@@ -0,0 +1,42 @@
+{{#list-item-text class="title fragments-title"}}
+    Fragments
+    {{#help-icon}}
+        Fragments
+        <p>
```
ruairif: The wording is a little bit technical maybe:

    A generated url is made up of one or more fragments. A fragment can be a single value, a list of different values or a range of numbers.

    Using fragments you can create a large number of urls.

PR #515, opened 2016-06-27
Add samples to train page clustering on subsequest crawls

ruairif:


slybot/slybot/clustering.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class PersistentClusteringMiddleware(object):
+    def __init__(self, dir, reset=False, stats=None):
+        try:
+            import anydbm as _dbm
+        except ImportError:
+            import dbm as _dbm
+        self.dbmodule = _dbm
+        self.dir = dir
```
sagelliv: Avoid using `dir` to avoid confusion with the built-in `dir` function.

PR #517, opened 2016-06-27
PORTIA-539 Start URLs for Slybot

sagelliv:
In order to deal with legacy urls and start urls, I created the `StartUrlCollection` to centralize the conditional creation of legacy and start url logic in one place.

### DIFF
--- a/slybot/slybot/starturls/__init__.py
+++ b/slybot/slybot/starturls/__init__.py
@@ -1,6 +1,105 @@
-from .generator import UrlGenerator
+from collections import OrderedDict as ODict
+from itertools import chain, product
+
+from scrapy.utils.spider import arg_to_iter
+
+import six
+from six.moves.urllib_parse import urlparse
+
+from .fragment_generator import FragmentGenerator
+from .generator import IdentityGenerator, UrlGenerator
 
 
-class StartUrls():
-    def __call__(self, spec):
-        return spec
+class StartUrlCollection(object):
+    def __init__(self, start_urls, generators=None, generator_type='start_urls'):
+        self.generators = generators
+        self.generator_type = generator_type
+        self.start_urls = [self._url_type(url) for url in start_urls]
+
+    def __iter__(self):
+        generated = (self._generate_urls(url) for url in self.start_urls)
+        for url in chain(*(arg_to_iter(g) for g in generated)):
+            yield url
+
+    def uniq(self):
+        return list(ODict([(s.key, s.spec) for s in self.start_urls]).values())
+
+    @property
+    def allowed_domains(self):
+        domains = [start_url.allowed_domains for start_url in self.start_urls]
+        return list(set(chain(*domains)))
+
+    def _generate_urls(self, start_url):
+        generator = self.generators[start_url.generator_type]
+        return generator(start_url.generator_value)
+
+    def _url_type(self, start_url):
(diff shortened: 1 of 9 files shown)


slybot/slybot/starturls/__init__.py
```diff
@@ -1,6 +1,101 @@
...
+
+from .fragment_generator import FragmentGenerator
+from .generator import IdentityGenerator, UrlGenerator
+
+
+class StartUrlCollection(object):
+    def __init__(self, start_urls, generators=None, generator_type='start_urls'):
+        self.generators = generators
+        self.generator_type = generator_type
+        self.start_urls = map(self._url_type, start_urls)
```
ruairif: Can you change this to be python 3 compatible?

slybot/slybot/starturls/__init__.py
```diff
@@ -1,6 +1,101 @@
...
+        for url in chain(*(arg_to_iter(g) for g in generated)):
+            yield url
+
+    def uniq(self):
+        return list(ODict([(s.key, s.spec) for s in self.start_urls]).values())
+
+    @property
+    def allowed_domains(self):
+        domains = [start_url.allowed_domains for start_url in self.start_urls]
+        return list(reduce(set.union, domains, set()))
```
ruairif: For python 3 you would need to import reduce from functools.
You could also use:
```python
list(set(itertools.chain(*domains)))
```

slybot/slybot/starturls/__init__.py
```diff
@@ -1,6 +1,101 @@
...
+                return set()
+            fragments[0] = self._join_fragments(product(fragment, fragments[0]))
+        return set()
+
+    def _join_fragments(self, fragments):
+        return map(lambda (a, b): ''.join([a, b]), fragments)
+
+    def _has_domain(self, url):
+        methods = ['path', 'params', 'query', 'fragment']
+        return any(getattr(urlparse(url), method) != '' for method in methods)
```
ruairif: Can you just parse the url once? There is a caching mechanism in urlparse to avoid the expensive parsing process but it's probably better to just read it from a local variable

slyd/slyd/resources/models.py
```diff
@@ -133,11 +133,10 @@ class FieldSchema(SlydSchema):
     class Meta:
         type_ = 'fields'
 
-
 class SpiderSchema(SlydSchema):
     id = fields.Str(dump_only=True, load_from='name')
     name = fields.Str(load_from='id')
-    start_urls = fields.List(fields.Str(), default=[])
+    start_urls = fields.Raw()
```
ruairif: Can you move these `slyd` changes to the other PR?

slybot/slybot/starturls/__init__.py
```diff
@@ -54,27 +54,31 @@ def __init__(self, spec, generators):
     def allowed_domains(self):
         if self._has_fragments:
             return self._find_fragment_domains()
-        return set([self.spec['url']])
+        return [self.spec['url']]
 
     def _find_fragment_domains(self):
         generator = self.generators[self.generator_type]
-        fragments = generator.process_fragments(self.spec)
+        fragmentLists = generator.process_fragments(self.spec)
```
ruairif: Can you change this to fragment_lists and change the other camelcase variables too?

PR #519, opened 2016-06-29
Populate project relationship from manager

ruairif:
When copying a spider from portia 1.0 to portia 2.0 the project id was populated from the previous project rather than the new project.

### DIFF
--- a/slyd/slyd/resources/spiders.py
+++ b/slyd/slyd/resources/spiders.py
@@ -56,7 +56,8 @@ def update_spider(manager, spider_id, attributes):
         if 'samples' in spider:
             del spider['samples']
         response = SpiderSchema(context=context).dump(spider).data
-        spider = _populate_relationships(spider, spider_id)
+        spider = _populate_relationships(
+            manager.project_name, spider, spider_id)
         new_spider = SpiderSchema(context=context).dump(spider).data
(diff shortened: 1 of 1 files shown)


slyd/slyd/resources/spiders.py
```diff
@@ -56,7 +56,7 @@ def update_spider(manager, spider_id, attributes):
         if 'samples' in spider:
             del spider['samples']
         response = SpiderSchema(context=context).dump(spider).data
-        spider = _populate_relationships(spider, spider_id)
+        spider = _populate_relationships(manager, spider, spider_id)
```
sagelliv: You can pass `manager.project_name` to `_populate_relationships` as only the `project_name` is used in `_populate_relationships`.

slyd/slyd/resources/spiders.py
```diff
@@ -84,7 +84,7 @@ def _check_spider_attributes(attributes, include_defaults=False):
 def _load_spider(manager, spider_id, include_samples=False):
     spider = manager.spider_json(spider_id)
     spider['id'] = spider_id
-    spider = _populate_relationships(spider, spider_id)
+    spider = _populate_relationships(manager, spider, spider_id)
```
sagelliv: Same as above ^

PR #524, opened 2016-07-06
[WIP] PORTIA-494: JSON API endpoints for ORM

michalmo:


slyd/slyd/orm/base.py
```diff
@@ -0,0 +1,550 @@
...
+            if attrname == primary_key:
+                dependencies[attrname] = set()
+            elif isinstance(field, DependantField):
+                dependencies[attrname] = {primary_key, field.when}
+            else:
+                dependencies[attrname] = {primary_key}
+        try:
+            ordered_fields = toposort_flatten(dependencies)
+        except ValueError as e:
+            raise ImproperlyConfigured(e.message)
```
sagelliv: This portion of code checks dependencies are acyclic and created the `ordered_fields` to attach it the class right? If this is the case, can you refactor this into a helper function so that we have only `cls._ordered_fields = self._build_ordered_fields`?

slyd/slyd/orm/base.py
```diff
@@ -0,0 +1,550 @@
...
+                                         if isinstance(f, BaseRelationship))
+        cls.opts = ModelOpts(meta, cls)
+        cls.collection = type(name + 'Collection', (ModelCollection,), {
+            'model': cls
+        })
+
+        for attrname, field in iteritems(fields):
+            if attrname in file_fields:
+                file_schema_attrs[attrname] = field
+            field.contribute_to_class(cls, attrname)
```
sagelliv: Right now we are delegating a call to `settattr(cls, ...)` to `field` which handles descriptor construction using duck typing on `Field` and `List` (fields.py). However, I think that `setattr` should be called in the meta class as the main responsibility of the meta class is this kind of "surgery" before a class is instantiated. A perhaps better alternative is to define a method which will build a descriptor based on the `attrname`:
```
class Field(fields.Field):
    def build_descriptor(self, attrname):
        return FieldDescriptor(attrname, self)

class List(fields.List, Field):
    def build_descriptor(self, attrname):
        return ListDescriptor(attrname=attrname)
```
and then replace `field.contribute_to_class(cls, attrname)` with `setattr(cls, attrname, field.build_descriptor(attrname))`.

slyd/slyd/jsonapi/schema.py
```diff
@@ -0,0 +1,216 @@
...
+                            reverse=reverse)
+
+        return models
+
+    @post_dump(pass_many=True)
+    def format_json_api_response(self, data, many):
+        response = super(JsonApiSchema, self).format_json_api_response(
+            data, many)
+        if 'included' in response:
+            response['included'].sort(key=itemgetter('id'))
```
ruairif: These 2 sorts can be done as a single sort using:
```python
response['included'].sort(key=itemgetter('type', 'id'))
```

portiaui/app/adapters/application.js
```diff
@@ -1,3 +1,400 @@
...
+                    return resolver.promise;
+                }
+            } else {
+                promise = this.ajax(url, method, request);
+            }
+
+            // handle our custom json api extension for listing records that were
+            // deleted in the backend and should be unloaded.
+            promise.then(response => {
+                if (response && response.links && response.links.profile &&
```
ruairif: Same logic as the UPDATES_EXTENSION above just with a different constant.

Does it make sense to create a separate function for this?

slyd/slyd/orm/models.py
```diff
@@ -0,0 +1,585 @@
...
+    class Meta:
+        path = u'items.json'
+        owner = 'project'
+        envelope = True
+        envelope_remove_key = True
+
+    @pre_load
+    def name_from_id(self, data):
+        if 'name' not in data:
+            # display_name ?
```
ruairif: Can be ignored. It was never used in production (portia-beta)

slyd/slyd/orm/models.py
```diff
@@ -0,0 +1,585 @@
...
+
+class Spider(Model):
+    # TODO: validate id against allowed file name
+    id = String(primary_key=True)
+    start_urls = List(Url)
+    # TODO: generated urls
+    links_to_follow = String(default='all', validate=OneOf(
+        ['none', 'patterns', 'all', 'auto']))
+    # TODO: compute automatically from start urls
+    allowed_domains = List(Domain)
```
ruairif: allowed_domains is already generated from start_urls (and sample urls) when a spider is run. This feature could be useful if a user wants to define custom allowed domains.

slyd/slyd/orm/models.py
```diff
@@ -0,0 +1,585 @@
...
+                        'data': {
+                            data_id: annotation_data,
+                        }
+                    }))
+
+        for item in items:
+            container_id = item.get('container_id')
+            if not container_id:
+                continue
+            if 'repeated' in item and item.pop('repeated'):
```
ruairif: This can be just `item.pop('repeated', None)` or `item.pop('repeated', False)`

slyd/slyd/orm/models.py
```diff
@@ -0,0 +1,585 @@
...
+    repo.set_db_url(rs['location'])
+    connection = repo.connection_pool.connectionFactory(repo.connection_pool)
+
+    repo = Repoman.open_repo('2222238', connection, author='michal <>')
+    storage = GitStorage(repo, branch='staff')
+    return storage
+
+
+if __name__ == '__main__':
+    storage = init_storage()
```
ruairif: Is this debug code?

slyd/slyd/jsonapi/schema.py
```diff
@@ -0,0 +1,216 @@
...
+
+            only = list(field_set | {'id'})
+            for name in relationship_names:
+                relationship = self._declared_fields[name]
+                if (name in relationship_set or relationship.related_url or
+                        relationship.self_url):
+                    only.append(name)
+            only.append('_url')
+
+        self.field_order = list(fields or []) + map(self.inflect, field_names)
```
ruairif: Can you use `list(map(...))` here for py3 compatibility

slyd/slyd/orm/base.py
```diff
@@ -0,0 +1,550 @@
...
+        })
+        cls.file_schema = type(cls.__name__ + 'Schema', (BaseFileSchema,),
+                               file_schema_attrs)
+
+        # add new model to registry by name
+        models[name] = cls
+        return cls
+
+    @property
+    def _file_model(cls):
```
sagelliv: Does this method have to be in the meta class?

slyd/slyd/resources/projects.py
```diff
@@ -1,144 +1,230 @@
...
+        project_manager = self.project_manager
+        project_id = self.args.get('project_id')
+        spider_id = self.args.get('spider_id', None)
+        spider_ids = [spider_id] if spider_id is not None else '*'
+        fmt = self.query.get('format', ['spec'])[0]
+        return ProjectDownloadResponse(
+            project_id, spider_ids, fmt, project_manager)
+
+
+class ProjectDataMixin(object):
```
sagelliv: I think it makes sense to have mixins for the different CRUD verbs for routes as these mixins need to have available much of the state of the object. However, in this case, the `ProjectDataMixin` only accesses two instance variables `self.request.auth_info` and `self.project_manager.all_projects`. We can have a class which accepts the `auth_info` and `projects` variables to its constructor and use composition instead of inheritance (mixin) for it to be used in the `ProjectRoute` and `SpiderRoute`.

slyd/slyd/resources/projects.py
```diff
@@ -1,144 +1,230 @@
...
-        'args': [project_id, spider_ids]
-    }
-    fmt = attributes['arguments'].get('format', ['spec'])[0]
-    file_content = manager.pm.download_project(project_id, spider_ids, fmt=fmt)
-    return ProjectsManagerFileResponse(file_content, command, manager.pm)
+class ProjectDownloadMixin(object):
+    @classmethod
+    def get_resources(cls):
+        for resouce in super(ProjectDownloadMixin, cls).get_resources():
+            yield resouce
```
sagelliv: Small typo: `resouce` -> `resource`

sagelliv:
`ProjectSchema`, `FieldSchema`, `SchemaSchema`, ... should be renamed to `ProjectSerializer`, `FieldSerializer`, etc. (in slyd/resources/serializers.py). The same change should be made to the parents `JsonApiSchema`, `JsonApiSchemaOpts`, `JsonApiSchemaMeta` in `slyd/jsonapi/schema.py`. This distinguishes the concept from the json schema used in slybot for validations, the model `Schema` used in portia-dash, and additionally is more explicit about the objects used as serializers.

slyd/slyd/resources/annotations.py
```diff
@@ -1,216 +1,88 @@
...
+
+from .route import (JsonApiRoute, ListModelMixin, RetrieveModelMixin,
+                    CreateModelMixin, UpdateModelMixin, DestroyModelMixin)
+from .serializers import AnnotationSchema
+from ..orm.base import AUTO_PK
+from ..orm.models import Project, Field, Item
+
+
+class AnnotationRoute(JsonApiRoute, ListModelMixin, RetrieveModelMixin,
+                      CreateModelMixin, UpdateModelMixin, DestroyModelMixin):
```
sagelliv: The `AnnotationRoute`, `SchemaRoute`, `ExtractorRoute` and others have this list of inheritance/mixins. We could make an aggregate class which captures these classes which go together:
```
class RestfulRoute(JsonApiRoute, ListModelMixin, RetrieveModelMixin,
                   CreateModelMixin, UpdateModelMixin, DestroyModelMixin):
    pass
```

slyd/slyd/orm/base.py
michalmo: i've refactored this to allow fields to specify their dependencies, and replaced this with one line in the metaclass. let me know if you still feel it warrants moving to a method.

slyd/slyd/orm/base.py
michalmo: this approach allows the field to do more than just set a single attributes on the class, eg. setting more attributes. right now relationships check that the relationship is correctly created in both directions. we don't have more advanced use cases yet but i think it's a good extension point to keep.
also it separates concerns. the field is responsible for knowing what it needs to do to add itself to the class.

portiaui/app/adapters/application.js
michalmo: refactored into a _getExtentionAliases method and added an extra usage.
it still loops over the aliases and reads from response.meta, but it's only a coincidence that both extensions add data to the top level meta, they don't need to, so i didn't want to add the reading in the helper too. is that ok?

slyd/slyd/orm/base.py
michalmo: python's property doesn't work on the class object, getting the attribute returns the property descriptor. but i've refactored this adding a custom class_property descriptor that allows the property to be read both from the class and instance, so it's better now.

slyd/slyd/resources/projects.py
michalmo: that would change the interface by which the routes are initialized in `slyd.api.APIResource`, and i'm not convinced that making that more complex to simplify this makes sense.
on the other hand making a `BaseProjectRoute` used by both `ProjectRoute` and `SpiderRoute` that includes the property or sets the value in `__init__` is fundamentally no different than this mixin (the subclass case just changes the bases of the classes but everything else remains the same).
or did you have some other refactoring in mind?

slyd/slyd/jsonapi/registry.py
```diff
@@ -0,0 +1,16 @@
+from slyd.orm.exceptions import ImproperlyConfigured
+
+
+__all__ = [
+    'schema',
+]
+
+schemas = {}
```
sagelliv: Can we change `schemas` to `serializers` to preserve the semantics of how this registry is used? (As we moved from calling `ObjectSchema` to `ObjectSerializer`)

slyd/slyd/jsonapi/registry.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_schema(schema_type):
```
sagelliv: Change this to `get_serializer` or `registry_serializer` for better indicating we are caching serializers in the registry.

slyd/slyd/jsonapi/serializers.py
```diff
@@ -0,0 +1,655 @@
...
+                if not rel_many:
+                    rel_options['allow_none'] = True
+                if field.polymorphic:
+                    schema_attrs[attrname] = PolymorphicRelationship(
+                        **rel_options)
+                else:
+                    schema_attrs[attrname] = Relationship(
+                        type_=type_from_model_name(field.model.__name__),
+                        id_field='pk',
+                        **rel_options)
```
sagelliv: Why do we have to copy over the entire `model._fields` to the serializer class? I think delegating to the model in the serializer is ok as we have a reference to it from the serializer `Meta`. If the issue is with the `links` attribute in the serializer we can make a factory method which will create the appropriate relationships with the relevant relationships options.

slyd/slyd/jsonapi/serializers.py
```diff
@@ -0,0 +1,655 @@
...
+        model = instance.__class__
+
+        errors = {}
+        fields = []
+        for attrname in model._ordered_fields:
+            if attrname in validated_data:
+                try:
+                    value = validated_data[attrname]
+                    if attrname in model._field_names:
+                        setattr(instance, attrname, value)
```
sagelliv: Perhaps a better way to model the interface for a model is to expose and `update` method which will store all the changes we wish to make. Once the `save` method is invoked, it will do a `save(only=fields)` call where the fields are the various `update` calls which were made before calling `save`.

slyd/slyd/jsonapi/serializers.py
michalmo: We're creating a new marshmallow_jsonapi schema here using the fields introspected from the model. The marshmallow.Schema metaclasses then further proces this input, and creates it's own structures (self._declared_fields and populating self.fields when called, etc.). Rather than try to implement that internal api, and keep it up to date if the library changes, i think it's simpler and safer to provide the fields as attributes in the class definition according to the documented public api. Although here we pass them directly to the marshmallow meta class, that's just regular python semantics.

sagelliv:
:+1:

PR #530, opened 2016-07-12
Improve repeated items for slybot

ruairif:
Extract repeated lists of values
Container can only extract a single item unless it wraps a repeated container
Better handle cases where not all repeated items have all fields
Fixed printing error with BaseExtractor
Removed legacy support (legacy now uses scrapely.InstanceBasedLearningExtractor)
Log slybot version when using spidermanager

### DIFF
--- a/slybot/slybot/plugins/scrapely_annotations/builder.py
+++ b/slybot/slybot/plugins/scrapely_annotations/builder.py
@@ -8,6 +8,7 @@ from itertools import tee, count, chain, groupby
 from operator import itemgetter
 from uuid import uuid4
 
+from .migration import _get_parent
 from .utils import (serialize_tag, add_tagids, remove_tagids, TAGID,
                     OPEN_TAG, CLOSE_TAG, UNPAIRED_TAG, GENERATEDTAGID)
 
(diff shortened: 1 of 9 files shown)


slybot/slybot/plugins/scrapely_annotations/builder.py
```diff
@@ -308,12 +309,33 @@ def apply_selector_annotations(annotations, target_page):
...
+                    'id': container_id,
+                    'annotations': {'#portia-content': '#dummy'},
+                    'text-content': '#portia-content',
+                    'container_id': annotation['container_id'],
+                    'field': field,
+                    'tagid': parent.attrib.get('data-tagid')
+                })
+                annotation['item_container'] = True
+                annotation['field'] = field
+                annotation['container_id'] = container_id
```
sagelliv: Are we mutating the `annotations` that are passed into this function? If so, I think we should separate the mutation from returning the `converted_annotations`.

slybot/slybot/plugins/scrapely_annotations/builder.py
```diff
@@ -308,12 +309,33 @@ def apply_selector_annotations(annotations, target_page):
...
+            continue
+
+        tagids = [int(e.attrib.get('data-tagid', 1e9)) for e in elems]
+        tagid = min(tagids)
+        if tagid is not None:
+            annotation['tagid'] = tagid
+            converted_annotations.append(annotation)
+        if (annotation.get('repeated') and
+                not annotation.get('item_container') and
+                len(annotation.get('annotations')) == 1):
```
sagelliv: Can you add a local variable describing what this conditional represents? I'm not sure what this case corresponds to so the variable would reveal the intent here.

slybot/slybot/plugins/scrapely_annotations/extraction.py
```diff
@@ -151,6 +151,21 @@ def __init__(self, annotation, attribute_descriptors=None):
     def _create_basic_extractor(cls, annotation, attribute_descriptors):
         return cls(annotation, attribute_descriptors)
 
+    def __str__(self):
+        messages = ['BasicTypeExtractor(']
+        annotation = self.annotation
+        if annotation.surrounds_attribute:
+            messages.append(self.annotation.surrounds_attribute)
```
sagelliv: `messages.append(annotation.surrounds_attribute or '')`

slybot/slybot/plugins/scrapely_annotations/builder.py
ruairif: They are meant to be mutated. This function used to just add the `tagid` key to the annotations so they would be applied to the correct elements. Now it also adds support for building a `RepeatedFieldsExtractor` by converting a repeated annotation to a container and a repeated field container.

Here's a quick overview of the 3 different types of containers:
**Container:** `item_container == True`.
**RepeatedContainer:** `item_container == True and repeated == True`.
**RepeatedFieldsContainer:** `item_container == True and repeated == True and field == True`.

PR #531, opened 2016-07-12
Repeated fields support UI

ruairif:
Depends on: https://github.com/scrapinghub/portia/pull/563

### DIFF
--- a/portiaui/app/components/annotation-options.js
+++ b/portiaui/app/components/annotation-options.js
@@ -3,6 +3,7 @@ import { getAttributeList } from './inspector-panel';
 
 export default Ember.Component.extend({
     uiState: Ember.inject.service(),
+    dispatcher: Ember.inject.service(),
 
     tagName: '',
 
@@ -57,7 +58,7 @@ export default Ember.Component.extend({
                 return this.get('cssSelector');
             } else {
                 const annotation = this.get('annotation');
-                annotation.setSelector(value);
+                this.setAnnotationSelector(annotation, value);
                 return value;
             }
         }
@@ -71,7 +72,7 @@ export default Ember.Component.extend({
             const annotation = this.get('annotation');
             if (value === null) {
                 this.set('invalidSelector', false);
-                annotation.setSelector(this.get('cssSelector'));
+                this.setAnnotationSelector(annotation, this.get('cssSelector'));
             } else {
                 let invalidSelector = false;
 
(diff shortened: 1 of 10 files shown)


portiaui/app/services/annotation-structure.js
```diff
@@ -146,14 +147,39 @@ const ElementStructure = Ember.Object.extend({
...
+                                        }
+                                        if (i > 1) {
+                                            break;
+                                        }
+                                    }
+                                    if (i > 1) {
+                                        annotation.set('repeated', true);
+                                        break;
+                                    }
+                                }
```
sagelliv: This is more inefficient but is a bit more readable.

```
const isRepeated = repeatedContainers.map((container) => {
    return elements.filter((child) => { container.contains(child); });
}).any((repeated) => { repeated.length > 1 });

if (isRepeated) {
    annotation.set('repeated', true);
}
```

michalmo: that will give a jshint warning about functions in function :(

portiaui/app/services/annotation-structure.js
```diff
@@ -146,14 +147,39 @@ const ElementStructure = Ember.Object.extend({
...
                             xpath: annotation.get('selectorGenerator.xpath')
                         });
                     }
+                    annotation.set('repeated', false);
                     if (selector) {
                         selectorMatcher.register(selector, setElements);
+                        const elements = selectorMatcher.query(selector);
+                        if (repeatedSelector && elements && elements.length > 1) {
+                            const repeatedContainers = annotation.get('containerGenerator.parent.repeatedContainers');
+                            if (repeatedContainers.length > 1) {
```
michalmo: this logic belongs in the selector generator. and we should just read repeated from it here

portiaui/app/services/annotation-structure.js
sagelliv: Extract the inner function out?

`repeatedContainers.map(containedChildren).any((repeated) => { repeated.length > 1 });`

sagelliv:
Rebase @ruairif

portiaui/app/components/annotation-options.js
```diff
@@ -96,9 +97,42 @@ export default Ember.Component.extend({
...
+                return (this._updateSelectorPromise = this._updateSelector(sample));
+            });
+        }
+
+        return this._updateSelectorPromise;
+    },
+
+    _updateSelector(sample) {
+        const dispatcher = this.get('dispatcher');
+        const promise = dispatcher.updateSampleSelectors(sample).then(() => {
```
sagelliv: Why not return here directly?

ruairif: The promise needs to be in the closure that wraps the resolving function

PR #532, opened 2016-07-13
PORTIA-551 JobQ integration with Portia

sagelliv:
Integrate jobQ running jobs into Portia UI with an ember service that can be injected throughout the application to query jobQ for running jobs.

### DIFF
--- a/portiaui/app/components/jobq-badge.js
+++ b/portiaui/app/components/jobq-badge.js
@@ -1,0 +1,18 @@
+import Ember from 'ember';
+const { computed } = Ember;
+
+export default Ember.Component.extend({
+    jobQ: Ember.inject.service(),
+
+    jobs: computed.alias('jobQ.jobs'),
+
+    jobCount: computed('jobs.count', function() {
+        return this.get('jobs').countForSpider(this.get('spider.id'));
+    }),
+
+    isActive: computed.gt('jobCount', 0),
+
+    badgeStyle: computed('jobCount', function() {
+        return Ember.String.htmlSafe('background-color: #d9534f');
+    })
+});
--- a/portiaui/app/components/list-item-icon.js
+++ b/portiaui/app/components/list-item-icon.js
@@ -1,5 +1,7 @@
 import IconButton from './icon-button';
 
 export default IconButton.extend({
-    classNames: ['list-item-icon']
+    classNames: ['list-item-icon'],
+    classNameBindings: ['hasIndent'],
+    hasIndent: false
 });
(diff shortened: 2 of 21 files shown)


portiaui/app/components/jobq-badge.js
```diff
@@ -0,0 +1,18 @@
...
+    jobs: computed.alias('jobQ.jobs'),
+
+    jobCount: computed('jobs.count', function() {
+        return this.get('jobs').countForSpider(this.get('spider.id'));
+    }),
+
+    isActive: computed.gt('jobCount', 0),
+
+    badgeStyle: computed('jobCount', function() {
+        return Ember.String.htmlSafe('background-color: #d9534f');
```
michalmo: can we set this using a class

portiaui/app/routes/projects/project/spider.js
```diff
@@ -1,24 +1,27 @@
...
+            spider_project_id: this.projectId(),
             spider_id: model.get('id')
         });
     },
 
-    setupController(controller) {
+    setupController(controller, model) {
         this._super(...arguments);
+
+        this.get('jobQ').start(this.projectId(), model.get('id'));
```
michalmo: i don't see stop called anywhere

portiaui/app/components/jobq-badge.js
```diff
@@ -0,0 +1,18 @@
+import Ember from 'ember';
+const { computed } = Ember;
+
+export default Ember.Component.extend({
+    jobQ: Ember.inject.service(),
+
+    jobs: computed.alias('jobQ.jobs'),
+
+    jobCount: computed('jobs.count', function() {
```
michalmo: will recomputing only when count changes be enough to catch all changes? what if one spider ends and another starts between polls

portiaui/app/helpers/has-job.js
```diff
@@ -0,0 +1,14 @@
...
+export default Ember.Helper.extend({
+    jobQ: Ember.inject.service(),
+
+    onJobCountChange: Ember.observer('jobQ.jobs.count', function() {
+        this.recompute();
+    }),
+
+    compute(params) {
+        const spider = params[0];
+        return this.get('jobQ.jobs').countForSpider(spider.get('id')) > 0;
```
michalmo: this can be called often. maybe it's better for the jobQ service to have a method that returns an Ember.Object with it's own computed properties just for that spider id

portiaui/app/components/list-item-icon.js
```diff
@@ -1,5 +1,7 @@
 import IconButton from './icon-button';
 
 export default IconButton.extend({
-    classNames: ['list-item-icon']
+    classNames: ['list-item-icon'],
+    classNameBindings: ['hasIndent'],
+    hasIndent: false
```
michalmo: i think that instead of the icon being responsible for indent we should have a spacer component that is used in the template when there is no badge.

PR #536, opened 2016-07-20
Schedule a spider to run from portia.

ruairif:
[Depends on ORM](https://github.com/scrapinghub/portia/pull/524)

Add button to schedule crawl from the UI
Add `schedule` and `_schedule_data` methods to projectspec for scrapyd scheduler
Add support for getting objects from mysql backend with partial ids
Add support for downloading individual branch or individual commit of project

Button in spider list
Button on hover in opened spider

### DIFF
--- a/portiaui/app/components/project-structure-listing.js
+++ b/portiaui/app/components/project-structure-listing.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import {computedCanAddSpider} from '../services/dispatcher';
 
 export default Ember.Component.extend({
+    api: Ember.inject.service(),
     browser: Ember.inject.service(),
     dispatcher: Ember.inject.service(),
     uiState: Ember.inject.service(),
@@ -43,6 +44,22 @@ export default Ember.Component.extend({
 
         removeSpider(spider) {
             this.get('dispatcher').removeSpider(spider);
+        },
+
+        runSpider(spider) {
+            this.get('api').post('schedule', {
+                model: spider,
+                jsonData: {data: {type: 'spiders', id: spider.id}}
+            }).then(data => {
+                this.get('notificationManager').showNotification(
+                    'Your spider has been scheduled successfully');
+            }, data => {
+                let error = data.errors[0];
+                if (error.status > 499) {
+                    throw data;
+                }
+                this.get('notificationManager').showNotification(error.title, error.detail);
+            });
         },
 
         validateSpiderName(name) {
--- a/portiaui/app/components/project-structure-listing.js
+++ b/portiaui/app/components/project-structure-listing.js
@@ -2,7 +2,6 @@ import Ember from 'ember';
 import {computedCanAddSpider} from '../services/dispatcher';
 
 export default Ember.Component.extend({
-    api: Ember.inject.service(),
     browser: Ember.inject.service(),
     dispatcher: Ember.inject.service(),
     uiState: Ember.inject.service(),
@@ -44,22 +43,6 @@ export default Ember.Component.extend({
 
         removeSpider(spider) {
             this.get('dispatcher').removeSpider(spider);
-        },
-
-        runSpider(spider) {
-            this.get('api').post('schedule', {
-                model: spider,
-                jsonData: {data: {type: 'spiders', id: spider.id}}
-            }).then(data => {
-                this.get('notificationManager').showNotification(
-                    'Your spider has been scheduled successfully');
-            }, data => {
-                let error = data.errors[0];
-                if (error.status > 499) {
-                    throw data;
-                }
-                this.get('notificationManager').showNotification(error.title, error.detail);
-            });
         },
 
         validateSpiderName(name) {
--- a/portiaui/app/components/spider-message.js
+++ b/portiaui/app/components/spider-message.js
@@ -1,0 +1,26 @@
+import Ember from 'ember';
+const { computed } = Ember;
+
+export default Ember.Component.extend({
+    api: Ember.inject.service(),
+    notificationManager: Ember.inject.service(),
+    hasSpider: computed.bool('currentSpider'),
+
+    actions: {
+        runSpider(spider) {
+            this.get('api').post('schedule', {
+                model: spider,
+                jsonData: {data: {type: 'spiders', id: spider.id}}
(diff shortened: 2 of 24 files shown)


slyd/slyd/gitstorage/projects.py
```diff
@@ -181,8 +198,12 @@ def copy_data(self, source, destination, spiders, items):
         copier = GitSpiderCopier(source, destination, branch)
         return json.dumps(copier.copy(spiders, items))
 
-    def download_project(self, name, spiders=None, version=None, fmt=None):
-        self._open_repo(name, read_only=True)
+    def download_project(self, name, spiders=None, version=None, fmt=None,
+                         branch=None, **kwargs):
```
sagelliv: Are you using `**kwargs` somewhere?

slyd/slyd/resources/spiders.py
```diff
@@ -11,8 +13,32 @@ class SpiderRoute(ProjectDownloadMixin, JsonApiRoute, ProjectDataMixin,
     list_path = 'projects/{project_id}/spiders'
     detail_path = 'projects/{project_id}/spiders/{spider_id}'
     download_path = 'projects/{project_id}/download/{spider_id}'
+    schedule_path = 'projects/{project_id}/spiders/{spider_id}/schedule'
     serializer_class = SpiderSchema
 
+    @classmethod
+    def get_resources(cls):
+        for resouce in super(SpiderRoute, cls).get_resources():
+            yield resouce
```
sagelliv: typo -> resource

slyd/slyd/gitstorage/projects.py
ruairif: Here's the spec for downloading from the FS storage implementation
https://github.com/scrapinghub/portia/pull/536/files#diff-5f33b932b8cf1d9581ad6fefd4f676f3R237
Here's where the download gets called:
https://github.com/scrapinghub/portia/pull/536/files#diff-909c4e43c52fe1edc233a4b28ff41609R97

The Git backend takes optional `branch` and `version` arguments but the FS backend does not. If we were to add an arg to the FS backend that is not used by the Git backend and there was no kwargs there then it would throw a TypeError because it got an unexpected additional argument.

So to answer your question it is not used but is there for consistency between the backends.

portiaui/app/components/project-structure-listing.js
```diff
@@ -24,6 +25,8 @@ export default Ember.Component.extend({
         }
     }),
 
+    runSpiderTooltipText: 'Run this spider.',
```
michalmo: can we keep the tooltip text in the template?

slyd/slyd/gitstorage/repo.py
```diff
@@ -177,7 +177,8 @@ class MysqlObjectStore(BaseObjectStore):
     statements = {
         "HAS": "SELECT EXISTS(SELECT 1 FROM objs WHERE `oid`=%s AND `repo`=%s)",
         "ALL": "SELECT `oid` FROM objs WHERE `repo`=%s",
-        "GET": "SELECT `type`, UNCOMPRESS(`data`) FROM objs WHERE `oid`=%s AND `repo`=%s",
+        "GET": "SELECT `type`, UNCOMPRESS(`data`), `oid` FROM objs WHERE `oid`=%s AND `repo`=%s",
+        "GET_SHORT": "SELECT `type`, UNCOMPRESS(`data`), `oid` FROM objs WHERE `oid`>=%s AND `oid` < %s AND `repo`=%s",
```
michalmo: I'd go with
```sql
SELECT `type`, UNCOMPRESS(`data`), `oid` FROM objs WHERE `oid` LIKE CONCAT(%s, '%%') AND `repo`=%s LIMIT 1
```

slyd/slyd/resources/spiders.py
```diff
@@ -1,18 +1,48 @@
...
+from .serializers import SpiderSchema
+from ..orm.models import Project, Spiders
 
 
 class SpiderRoute(ProjectDownloadMixin, ProjectDataMixin, JsonApiModelRoute):
     list_path = 'projects/{project_id}/spiders'
     detail_path = 'projects/{project_id}/spiders/{spider_id}'
     download_path = 'projects/{project_id}/download/{spider_id}'
+    schedule_path = 'projects/{project_id}/spiders/{spider_id}/schedule'
+    serializer_class = SpiderSchema
```
michalmo: serializer_class is no longer used after the last refactoring

portiaui/app/components/project-structure-listing.js
```diff
@@ -45,6 +48,22 @@ export default Ember.Component.extend({
             this.get('dispatcher').removeSpider(spider);
         },
 
+        runSpider(spider) {
+            this.get('api').put('schedule', {
```
michalmo: shouldn't this be a post?

slyd/slyd/gitstorage/projects.py
sagelliv: :+1:

sagelliv:
@ruairif: use the JobQ components for running a spider and do not use it in the spider listing.

portiaui/app/components/spider-message.js
```diff
@@ -0,0 +1,26 @@
...
+const { computed } = Ember;
+
+export default Ember.Component.extend({
+    api: Ember.inject.service(),
+    notificationManager: Ember.inject.service(),
+    hasSpider: computed.bool('currentSpider'),
+
+    actions: {
+        runSpider(spider) {
+            this.get('api').post('schedule', {
```
sagelliv: I think it makes more sense to have a function in the `api` service which receives a spider and takes care of the interactions with the `notificationManager`.

ruairif: That api service is an external library, `ember-api-requests`. We could create our own api service that uses this but so far it is only used here.

slyd/slyd/splash/utils.py
```diff
@@ -116,17 +116,17 @@ def _decode(html, default=None):
 
 
 def lazy_property(func):
```
michalmo: this can be removed in favor of `django.utils.functional.cached_property` which has a cleaner implementation

PR #537, opened 2016-07-20
Build splash assets during slybot build

ruairif:
Waiting to see if es5-shim can be removed

Overwrite bdist_egg and sdist to build the latest splash assets
Remove es5-shim from assets


slybot/slybot/splash-script-combined.js
```diff
@@ -2004,4 +465,5 @@ function performEvents(eventList, callback) {
 window.__slybot__performEvents = performEvents;
 
 })();
-})();
+
+})();
```
michalmo: is this file still needed in the package?

slyd/splash_utils/compile_slybot.sh
```diff
@@ -4,7 +4,6 @@
     echo ";(function(){"
 
     cat 'es5-shim-fixes.js'
-    cat '../node_modules/es5-shim/es5-shim.js'
 
     # Page actions scripts
     cat 'waitAsync.js'
```
michalmo: do we need this script if it gets build in the package?

PR #540, opened 2016-07-25
PORTIA-472: Explicit selector recomputation on user input

michalmo:


portiaui/app/utils/selectors.js
```diff
@@ -879,6 +879,71 @@ export function makeItemsFromGroups(groups) {
     return items;
 }
 
+function createSelectorGenerators(structure, selectorMatcher, accumulator) {
```
sagelliv: This is mutation `accumulator` and returns `generators`. Can we implement this without mutating a passed in parameter? If this is only used in `updateStructureSelectors` we don't need to mutate the `accumulator`, just construct it in the recursion by creating a helper function and then calling it in the invocation of `createSelectorGenerators`.

portiaui/app/services/annotation-structure.js
```diff
@@ -270,24 +176,59 @@ const ElementStructure = Ember.Object.extend({
...
+        structurePromise.then(structure => {
+            if (structurePromise === this.currentPromise) {
+                delete this.currentPromise;
+                this.set('definition', structure);
+            }
+        });
+    }))
 });
 
+export function createStructure(sample) {
```
sagelliv: Can you refactor some of the callbacks into separate functions to make this method a bit more abstract? The nesting here makes this method hard to understand.

portiaui/app/utils/selectors.js
```diff
@@ -879,6 +879,71 @@ export function makeItemsFromGroups(groups) {
...
+
+    return generators;
+}
+
+export function updateStructureSelectors(structure, selectorMatcher) {
+    const accumulator = [];
+    createSelectorGenerators(structure, selectorMatcher, accumulator);
+    for (let [annotation, selectorGenerator] of accumulator) {
+        const selector = selectorGenerator.get('selector');
+        if (selectorGenerator instanceof AnnotationSelectorGenerator) {
```
sagelliv: Can you replace these conditionals with polymorphism? Have the generator classes respond to a method which takes care of updating the structure given the annotation. In case the selector generator is not a `AnnotationSelectorGenerator` or `ContainerSelectorGenerator`, just do `this.destroy()`.

portiaui/app/services/annotation-structure.js
michalmo: this promise juggling is needed to get the mix of promises and PromiseManyArrays to resolve to a plain object at the end. if we split this up it will be even harder to see the structure that's expected as a result of this call.

portiaui/app/utils/selectors.js
michalmo: the selector generator by design only operates on a structure format and doesn't need to know about the base annotation types that created the structure (it could be annotations and items, but in the future it may be link annotations, actions etc.) so it's the responsibility of the api consumer that created the structure to know how to apply the generated selectors back onto the annotations the structure was created from. having sait that, this clear that this belongs in annotation-structure.js

sagelliv:
@michalmo rebase

sagelliv:
@michalmo: Rebasing/cherry picking is nontrivial, as @ruairif made a [related change](https://github.com/scrapinghub/portia/commit/e1d51caadfa7faf0f1044855e625757d371f7800) which when rebased causes problems with repeated annotations.

PR #541, opened 2016-07-25
PORTIA-415: Queue model updates

michalmo:


portiaui/app/models/base.js
```diff
@@ -3,8 +3,110 @@ import DS from 'ember-data';
...
+            // model has started saving again, wait for the next didCommit event
+            return;
+        }
+        if (nextModel.get('isDeleted') && !nextModel.get('hasDirtyAttributes')) {
+            // model has been deleted
+            continue;
+        }
+        if (nextModel.get('pendingDelete')) {
+            nextModel.deleteRecord();
+            nextModel.set('pendingDelete', false);
```
sagelliv: After deleting a record don't we want to `continue`? Is there a case when a record has `pendingDelete === true` and `pendingSave === true`? With `continue` it would be clearer if it is the case that deleting a record just continues to the next model in `modelActionQueue`.

portiaui/app/models/base.js
```diff
@@ -3,8 +3,110 @@ import DS from 'ember-data';
...
+}
+
+function mergeSaveOptions(dst, src) {
+    if (!dst) {
+        return src;
+    }
+
+    if (!src) {
+        src = {};
+    }
```
sagelliv: Compact: `src = src || {};`

portiaui/app/models/base.js
```diff
@@ -3,8 +3,110 @@ import DS from 'ember-data';
...
+
+    isDeleted: Ember.computed('currentState', 'pendingDelete', function() {
+        return get(this._internalModel.currentState, 'isDeleted') || this.get('pendingDelete');
+    }).readOnly(),
+
     save(options) {
+        if (currentActionModel && currentActionModel.get('isSaving') &&
+                // allow coalesced requests through since we're inside the main save call
+                !(options && options.adapterOptions && options.adapterOptions.coalesce &&
+                  options.adapterOptions.coalesce.type === 'extra')) {
```
sagelliv: Can you refactor this boolean into `savingAndNotExtra` or some other descriptive name of the boolean?

sagelliv:
:+1: After minor comments and rebase. @michalmo

portiaui/app/models/base.js
michalmo: deleteRecord marks it as deleted, destoryRecord calls deleteRecord followed by save.
we don't continue here so that we can proceed to call save if there is a pending save too.
Maybe this needs a comment.

PR #548, opened 2016-08-03
PORTIA-598 Normalize start urls

sagelliv:
The `StartUrlCollection` allows three types of start urls:

* `StringUrl`: legacy start url which is just a string
* `GeneratedUrl`: legacy start url which was used for kimono urls
* `StartUrl`: new start urls based on fragments

I exposed `normalize` in `StartUrlCollection` which normalizes all types of urls to use the new fragments structure. This allows us to normalize start urls in the `pre_load` hook in the ORM to start gradually converting start urls.

This normalization in turn simplifies the validations we perform on the start urls field for spiders using the ORM validators.

### DIFF
--- a/slybot/slybot/starturls/generated_url.py
+++ b/slybot/slybot/starturls/generated_url.py
@@ -32,7 +32,7 @@ class GeneratedUrl(object):
     @property
     def normalized_fragments(self):
         fixed, path = self._fixed_fragments, self._path_fragments
-        fragments = zip(fixed, path)
+        fragments = list(zip(fixed, path))
 
         # Missed last fixed fragment when using zip
         if len(fixed) == len(path) + 1:
@@ -64,8 +64,9 @@ class GeneratedUrl(object):
         template_params = self.spec['params_template']
 
         query_params = []
+        param_names = {p['name'] for p in params}
         for query, value in template_params:
-            if query not in [p['name'] for p in params]:
+            if query not in param_names:
                 query_params.append({
                     'name': query,
                     'type': 'default',
--- a/slybot/slybot/starturls/generated_url.py
+++ b/slybot/slybot/starturls/generated_url.py
@@ -1,0 +1,133 @@
+from itertools import chain
+
+
+class GeneratedUrl(object):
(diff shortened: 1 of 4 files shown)


slybot/slybot/starturls/generated_url.py
```diff
@@ -0,0 +1,133 @@
...
+        base_url = self.spec['template'].format(*paths)
+
+        query_params = [normalize_url_query_param(path, is_first=(i == 0))
+                        for i, path in enumerate(self._query_params)]
+        return base_url + ''.join(query_params)
+
+    @property
+    def normalized_fragments(self):
+        fixed, path = self._fixed_fragments, self._path_fragments
+        fragments = zip(fixed, path)
```
ruairif: Can you use `list(zip(fixed, path))` for python 3 compatability

slybot/slybot/starturls/generated_url.py
```diff
@@ -0,0 +1,133 @@
...
+        return list(chain(*fragments))
+
+    @property
+    def _query_params(self):
+        params = self.spec['params']
+        template_params = self.spec['params_template']
+
+        query_params = []
+        for query, value in template_params:
+            if query not in [p['name'] for p in params]:
```
ruairif: Could you use:
```python
param_names = {p['name'] for p in params}
...
    if query not in param_names:
        ...
```

slybot/slybot/starturls/generated_url.py
```diff
@@ -0,0 +1,133 @@
...
+    return [first_fragment, normalize_path(x)]
+
+
+def normalize_default(x):
+    return str(x['values'][0])
+
+
+def normalize_range(x):
+    a, b = x['values'][0:2]
+    b_inclusive = str(int(b) - 1)
```
ruairif: Should this be `str(int(b) + 1)`?

sagelliv: In the `_process_range` function in the `FragmentGenerator` we are adding 1 so we need to subtract here.

PR #562, opened 2016-08-26
Fix some JS hint errors

ruairif:


portiaui/app/components/input-with-clear.js
```diff
@@ -9,10 +9,10 @@ export default Ember.Component.extend({
     actions: {
         clear() {
             this.set('value', '');
-            this.attrs.clear()
+            this.attrs.clear();
```
sagelliv: Never use `attrs`: https://locks.svbtle.com/to-attrs-or-not-to-attrs. `this.get('clear')()`;

PR #563, opened 2016-08-26
Fix error when creating item with single repeated field

ruairif:
Needed for https://github.com/scrapinghub/portia/pull/531

Error occurs when creating a repeated field annotation as the only annotation for an item.

When this happens the repeated field splits into a field and surrounding container as designed but it is possible that the new container will overwrite the item container. This PR handles this by moving the item container to the next parent element if there is a conflict between the 2 containers.

### DIFF
--- a/slybot/slybot/plugins/scrapely_annotations/builder.py
+++ b/slybot/slybot/plugins/scrapely_annotations/builder.py
@@ -254,7 +254,7 @@ def _filter_annotations(annotations):
     selector, tagid = [], []
     for ann in annotations:
         if ann:
-            if ann.get('selector') or ann.get('accept_selectors'):
+            if ann.get('selector'):
                 selector.append(ann)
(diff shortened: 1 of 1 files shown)


slybot/slybot/plugins/scrapely_annotations/builder.py
```diff
@@ -294,53 +294,84 @@ def apply_selector_annotations(annotations, target_page):
...
     return converted_annotations
 
 
+def tagid_for_annotation(annotation, page):
+    selector = annotation.get('selector')
+    if not selector:
+        accepted_elements = set(
+            chain(*[[elem._root for elem in page.css(sel)]
+                    for sel in annotation.get('accept_selectors', [])
+                    if sel])
```
sagelliv: Refactor these chains? The only things that changes is `accept_selectors` vs `reject_selectors`

PR #564, opened 2016-08-26
Fix CI Build

sagelliv:
* Remove `portia/slyd/tests` and add `portia/slyd/orm/tests` to `slyd/tests`.
* Remove `npm test` from Travis to be replaced by tests in the `portia/portiaui` based ember app once it has non trivial tests.


sagelliv:
Fix ORM tests regarding `start_urls`, and a Sample's `body` and `rendered_body`.

sagelliv:
Instead of ensuring ordering in a `@post_dump` hook, opted for fixing the non-passing tests by changing the start_urls key orders.

PR #582, opened 2016-09-20
Django feed urls

sagelliv:
Added feed urls: a way for users to specify a url (drive, gist, dropbox, etc.) with newline separated urls for Portia to crawl.


slybot/slybot/starturls/feed_generator.py
```diff
@@ -0,0 +1,15 @@
+from scrapy import Request
```
ruairif: Can the slybot changes be moved to a different PR so they can be released on PyPi

slybot/slybot/starturls/__init__.py
```diff
@@ -6,16 +6,16 @@
 import six
 from six.moves.urllib_parse import urlparse
 
+from .feed_generator import FeedGenerator
```
ruairif: Is there a reason to import FeedGenerator?

sagelliv: So that we can have `from slybot.starturls import FeedGenerator`. The `__init__.py` file is used here to expose what we want from the python package.

sagelliv:
Depends on https://github.com/scrapinghub/portia/pull/587 and needs a rebase with `nui-develop` to work.

slybot/slybot/starturls/feed_generator.py
sagelliv: Done

portiaui/app/models/start-url.js
```diff
@@ -33,41 +67,48 @@ export default function startUrl(spec) {
...
-            base.fragments = urlObject.fragments;
+        if (url.includes('dropbox.com') && notRaw) {
+            return url + '&raw=1';
         }
-        return base;
-    }
 
-    let urlObject = buildUrlObject();
-    urlObject.isGenerated = urlObject.type === 'generated';
+        if (url.includes('google.com') && !url.includes('export')) {
```
sagelliv: Bitbucket snippets are tricky to get right. I opted for going with the url `http://files.scrapinghub.com/portia/urls.txt`.

PR #598, opened 2016-10-12
Explicitly handle CSS and XPath annotations in Slybot

ruairif:
- [x] Process extracted data using ItemProcessor and ItemField classes
- [x] Add tag ids to extracted html
- [x] Pass region tag ids of container to ItemProcessor
- [x] Extract CSS and XPath selectors before finalising extracted data


slybot/slybot/plugins/scrapely_annotations/extraction/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+from scrapely.extraction.pageobjects import TokenDict
+from scrapely.extraction.regionextract import (TraceExtractor,
+                                               TemplatePageExtractor)
+from scrapy.utils.spider import arg_to_iter
+
+from .container_extractors import BaseContainerExtractor, ContainerExtractor
+from .pageparsing import parse_template
+from .region_extractors import BaseExtractor
+from .utils import _count_annotations
+from ..processors import ItemProcessor
```
sagelliv: I would leave this `__init__.py` with only the imports of files we want to expose from this `extraction` package in order to do `from .extraction import SlybotIBLExtractor`. I'd move `TemplatePageMultiItemExtractor` and `SlybotIBLExtractor` to their own file. This way, `__init__.py` cleanly describes what can be found in the package and exposes it in a natural way.

slybot/slybot/plugins/scrapely_annotations/processors.py
```diff
@@ -9,72 +9,13 @@
...
-    pass
-
-
-_DEFAULT_EXTRACTOR = FieldTypeManager().type_processor_class(u'raw html')()
+from .utils import (cached_property, load_annotations, region_id,
+                    _DEFAULT_EXTRACTOR)
+from .exceptions import ItemNotValidError, MissingRequiredError
 
 
 def _compose(f, g):
```
sagelliv: Move to utils.

PR #600, opened 2016-10-13
Fragment validations

sagelliv:
Generated urls have validations in `portiaui` now. This was achieved using [changesets](https://github.com/DockYard/ember-changeset) and [concurrency](https://github.com/machty/ember-concurrency) to manage the error message which appears for the user. An observer had to used to control how we hide and show the tooltip but there might be an alternative implementation.

We now accept digits and letters in the `portia_server` as well.

[Demo](https://portia-generated-validation-demo.scrapinghub.com/)


sagelliv:
Added concurrency to saving a spider to avoid saving the spider before creating a new fragment. Changed the [triangle-error](https://github.com/scrapinghub/portia/pull/600/commits/5a833bfc7d2ff87c80a7aa0488a7412e3c5f5e52#diff-144ae962d620bb9e3af6191ec64fee36R12) . Is this what you were referring to @ruairif?

sagelliv:
Do not merge: have to fix a bug with the fragment type changing and flickering.

portiaui/app/validators/range.js
```diff
@@ -0,0 +1,43 @@
...
+function validateIncreasing(endpoints, isIncreasing) {
+    const [a, b] = endpoints.split('-');
+    const msg = `Range must be increasing. Try swapping to ${b}-${a}.`;
+    return isIncreasing(a, b) || msg;
+}
+
+export default function validateRange() {
+    return (key, newValue/*, oldValue, changes */) => {
+        if (isRangeIncomplete(newValue)) {
+            return 'Range is incomplete. It must have two endpoints.';
```
ruairif: `A range must have both a start and an end.`

portiaui/app/validators/range.js
```diff
@@ -0,0 +1,43 @@
...
+    return endpoints.split('-').every((x) => x.length === 1);
+}
+
+function isRangeIncomplete(endpoints) {
+    return endpoints.split('-').contains('');
+}
+
+function validateIncreasing(endpoints, isIncreasing) {
+    const [a, b] = endpoints.split('-');
+    const msg = `Range must be increasing. Try swapping to ${b}-${a}.`;
```
ruairif: `A range must be increasing. Try swapping to ${b}-${a}.`

Just to match the other errors

PR #608, opened 2016-10-24
Fix error where renaming spider causes it to overwrite another spider

ruairif:
Raise error in the schedule too

### DIFF
--- a/portiaui/app/components/project-structure-listing.js
+++ b/portiaui/app/components/project-structure-listing.js
@@ -53,11 +53,20 @@ export default Ember.Component.extend({
             this.get('dispatcher').removeSpider(spider);
         },
 
-        validateSpiderName(name) {
+        validateSpiderName(spider, name) {
             const nm = this.get('notificationManager');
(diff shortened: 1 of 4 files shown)


sagelliv:
Can you include tests for this and add the validation in the frontend as well? @ruairif

portiaui/app/components/project-structure-listing.js
```diff
@@ -53,13 +53,22 @@ export default Ember.Component.extend({
...
             const nm = this.get('notificationManager');
             if(!/^[a-zA-Z0-9][a-zA-Z0-9_\.-]*$/.test(name)) {
                 nm.showWarningNotification(`Invalid spider name.
                     Only letters, numbers, underscores, dashes and dots are allowed.`);
                 return false;
             }
+            if (spider.get('id') === name) {
+                return true;
+            }
+            const spiders = this.get('project.spiders').map((s) => s.get('id'));
```
sagelliv: Does `this.get('project.spiders').mapBy('id')` work?

PR #611, opened 2016-10-25
Send tag as surrounding region

ruairif:
Small tests clean up


slybot/slybot/tests/test_multiple_item_extraction.py
```diff
@@ -28,11 +26,9 @@
 from scrapely.extraction.pageparsing import parse_extraction_page
 from scrapely.htmlpage import HtmlTagType
 
+from .utils import (_open_spec, _open_sample_and_page, _spider, _PATH,
+                    _open_spider_page_and_results)
```
sagelliv: We should not be exposing "private" functions to be used in the tests. Can you change these to be without underscores as we are exposing them publicly?

PR #615, opened 2016-10-28
Encode page html before extraction if unicode. Save sample html if missing

ruairif:


slyd/slyd/splash/commands.py
```diff
@@ -45,11 +46,11 @@ def save_html(data, socket):
     if sample.rendered_body and not data.get('update'):
         return
     if 'use_live' not in data:
-        data['use_live'] = c.using_js
-    sample.body = 'rendered_body' if data.get('use_live') else 'original_body'
-    sample.rendered_body = c.html
-    sample.original_body = c.raw_html
-    _update_sample(data, socket, sample, save=True)
+        data['use_live'] = item_checker.using_js
+    sample.body = 'rendered_body' if data['use_live'] else 'original_body'
```
sagelliv: Isn't `data.get('use_live')` safer? What if the `data` structure has the old format?

ruairif: The 2 lines above this are:
```python
if 'use_live' not in data:
    data['use_live'] = item_checker.using_js
```
So it should be safe

PR #616, opened 2016-10-28
Add extract endpoint with asynchronous url loading

ruairif:


portia_server/portia_api/utils/extract.py
```diff
@@ -0,0 +1,109 @@
...
+    return defer.DeferredList(deferreds)
+
+
+@wait_for(timeout=50)
+def load_urls(urls):
+    return _load_urls(urls)
+
+
+class Pages(object):
+    def __init__(self, urls, user=None):
```
sagelliv: `user` is not used in this class. Will you need it?

portia_server/portia_api/utils/extract.py
```diff
@@ -0,0 +1,109 @@
...
+                results.append(self.process(*result))
+        if errors and not results:
+            raise FetchError(errors)
+        return results
+
+    def process(self, url, page):
+        return (url, HtmlResponse(url, body=page))
+
+
+def extract_items(spider, pages):
```
sagelliv: Can you extract this function to the `Pages` class? You can change the signature to
```
def extract_items(self, spider):
    ...
```
This way you have access to `this.fetch` in pages. You might also want to inject the `spider` in the constructor to have access to `this.spider` and remove the dependency from the method `extract_items` altogether.

portia_server/portia_api/resources/spiders.py
```diff
@@ -26,6 +28,23 @@ def get_collection(self):
...
 
     @detail_route(methods=['post'])
+    def extract(self, *args, **kwargs):
+        pages = self.page_class(self.data, self.user)
+        try:
+            instance = self.get_instance()
+            spider = load_spider(self.storage, instance)
+        except (ValueError, KeyError, IndexError):
+            import traceback
+            traceback.print_exc()
```
ruairif: Yes. I'm not sure if using all of those errors is too general. We may need to add custom errors when building slybot spiders

portia_server/portia_api/utils/extract.py
ruairif: It could be useful for custom implementations but I'll remove it and add a separate build_pages method

portia_server/portia_api/utils/extract.py
ruairif: :+1:

PR #619, opened 2016-11-01
QA Bug Fixes

ruairif:
Update extracted links when items are extracted
Limit item nesting limit. Prioritize showing nested item name over schema name

### DIFF
--- a/portiaui/app/components/list-item-add-annotation-menu.js
+++ b/portiaui/app/components/list-item-add-annotation-menu.js
@@ -7,6 +7,8 @@ export default Ember.Component.extend({
 
     item: null,
 
+    allowNesting: Ember.computed.lt('item.depth', 4), 
+
     actions: {
         addAnnotation() {
             const item = this.get('item');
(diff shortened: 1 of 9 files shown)


portiaui/app/components/list-item-add-annotation-menu.js
```diff
@@ -7,6 +7,10 @@ export default Ember.Component.extend({
 
     item: null,
 
+    allowNesting: Ember.computed('item.depth', function() {
```
sagelliv: `allowNesting: Ember.computed.lt('item.depth', 4),`

portiaui/app/models/item.js
```diff
@@ -40,5 +40,10 @@ export default BaseAnnotation.extend({
                         []
                 )
             )));
-        })
+        }),
+
+    depth: Ember.computed('parent', function() {
```
sagelliv: Should it depend on `parent.depth`?

PR #627, opened 2016-11-09
PORTIA-688: Schedule spider

sagelliv:


sagelliv:
`self.storage.checkout( ... )` mutates the `self.storage` instance. In case `branch` and `commit` are both `None`, this method will checkout the `master` branch and latest commit. In this case, the side effect of calling this method is setting the `self.storage._commit` to the latest commit of master as desired. @ruairif

PR #630, opened 2016-11-10
Handle nested items

ruairif:
Fix nested items using item processor
Add `built` flag to samples to avoid re processing samples


slybot/slybot/plugins/scrapely_annotations/processors.py
```diff
@@ -100,13 +100,8 @@ def _process_fields(self, data):
...
             if hasattr(field, 'fields'):
-                child = None
-                if len(field.fields) == 1:
-                    child = field.fields.values()[0]
-                if child and field.descriptor == child.extractor:
-                    fields[child.id] = child
-                else:
-                    fields[field.name] = field
+                _id, field = self._process_item(field)
+                fields[_id] = field
```
sagelliv: This would be cleaner if you do `self._add_field(field, fields)` and do the `fields[_id] = field` in the helper method.

slybot/slybot/plugins/scrapely_annotations/processors.py
```diff
@@ -465,3 +478,10 @@ def __repr__(self):
         return u'%s(%s, field=%s, extractor=%s, adaptors=%s)' % (
             self.__class__.__name__, str(self), self._field, self.extractor,
             self.adaptors)
+
+
+class ProcessedItemField(ItemField):
```
sagelliv: This can be accomplished with
```
class MetaField(ItemField):
    def _field_type(self):
       return META_FIELD
```
This would be the first step to break apart the `ItemField` class into the `DefaultField`, `ProcessingField` and `MetaField` subclasses.

PR #636, opened 2016-11-15
Change data-scrapy-annotate format from `{...}` to `[{...}, ...]`

ruairif:
**Advantages:**
- Allow arbitrary nesting
- Containers can now share an element
- Annotations and containers can now be applied to the same element
- Annotations can share an element and not belong to the same container


slybot/slybot/plugins/scrapely_annotations/extraction/pageparsing.py
```diff
@@ -25,6 +33,129 @@ def to_template(self, descriptors=None):
                                   self.html_page.page_id, self.ignored_regions,
                                   self.extra_required_attrs, descriptors)
 
+    @staticmethod
```
sagelliv: If we are including this much from the `TemplatePageParser` directly, why don't we copy it over entirely? This eases the burden of rebasing it on top of scrapely later on.

ruairif: Looking at the scrapely code I prefer to keep it like this. It's quite easy to see what has been overwritten and this could even be backported to scrapely as is.
It's also unlikely that there will be any change to the TemplateParser in scrapely so this won't be a problem anyway.

PR #638, opened 2016-11-18
Move html to new file in portia server to reduce sample file size

ruairif:


portia_server/portia_orm/relationships.py
```diff
@@ -89,6 +90,21 @@ def replace_collection(self, collection, values):
         collection.update(values)
 
 
+class HasOneDescriptor(BelongsToDescriptor, BaseRelationshipDescriptor):
```
sagelliv: Try to make this work by either implementing the try catch logic in the `BelongsToDescriptor` to make this syntactic sugar, or get rid of it altogether and use `BelongsTo` in the ORM.

PR #644, opened 2016-11-30
PORTIA-703: Unexpected double click error

sagelliv:
Use droppable tasks to enhance UI experience for selecting elements when annotating (when the overlay appears) and removing annotation fields.


portiaui/app/services/dispatcher.js
```diff
@@ -240,6 +241,12 @@ export default Ember.Service.extend({
     },
 
     saveAnnotationAndRelatedSelectors(annotation) {
+        if (!annotation.get('ownerSample')) {
+          return new Ember.RSVP.Promise(function(resolve) {
```
michalmo: you can use `return Ember.RSVP.resolve()`

PR #663, opened 2016-12-30
Fix migration errors when moving to new format

ruairif:


slybot/slybot/plugins/scrapely_annotations/migration.py
```diff
@@ -598,10 +603,11 @@ def find_generated_annotation(elem):
         if (node is not None and isinstance(node, _Element) and
                 node.tag.lower() == 'ins'):
             last_node_ins = True
+            text_len = len('' if node.text is None else node.text)
```
sagelliv: `text_len = len(node.text or '')`

slybot/slybot/starturls/generated_url.py
```diff
@@ -23,7 +23,18 @@ def normalized(self):
...
+        except IndexError:
+            # Fix templates that have more sections than path pieces
+            template = self.spec['template']
+            while '{}' in template:
+                template = self.spec['template'].strip('{}/')
+                try:
+                    base_url = template.format(*paths)
+                    break
+                except IndexError:
+                    continue
```
sagelliv: Can you add a test case for this?

ruairif: Done

PR #669, opened 2017-01-04
Add schema_id to newly created items if none exists

ruairif:


slybot/slybot/plugins/scrapely_annotations/migration.py
```diff
@@ -188,7 +190,6 @@ def port_sample(sample, schemas=None, extractors=None):
     # Update annotations
     sample['plugins']['annotations-plugin']['extracts'] = new_annotations
     sample['version'] = SLYBOT_VERSION
-    schema_id, schemas = guess_schema(sample, schemas)
     container['schema_id'] = schema_id
```
sagelliv: Can you move this by passing the schema_id on line 177?

slybot/slybot/plugins/scrapely_annotations/migration.py
```diff
@@ -401,7 +402,7 @@ def _get_highest(annotations, upto, sel):
 
 
 def _create_container(element, container_id, repeated=False, siblings=0,
-                      field=None, selector=None):
+                      field=None, selector=None, schema=None):
```
sagelliv: `schema_id=None` might be a clearer name.

Issue #648, opened 2016-12-09
Portia 2.0


kode-ninja:
Hi,

Which branch should I checkout to try Portia 2.0?

Thanks

sagelliv:
Hi @kode-ninja,

You can use the `nui-develop` branch.

kode-ninja:
Thanks.

It seems slyd fails to start when running "vagrant up". Any advice?

```
==> default: Installing collected packages: slybot
==> default:   Running setup.py develop for slybot
==> default:
==> default:     warning: no files found matching 'slybot/splash-script-combined.js'
==> default:     Creating /usr/local/lib/python2.7/dist-packages/slybot.egg-link (link to .)
==> default:     Adding slybot 0.13.0b28 to easy-install.pth file
==> default:     Installing slybot script to /usr/local/bin
==> default:     Installing portiacrawl script to /usr/local/bin
==> default:
==> default:     Installed /vagrant/slybot
==> default: Successfully installed slybot
==> default: Cleaning up...
==> default: Executing command: configure_nginx
==> default: Executing command: configure_initctl
==> default: Starting slyd service
==> default: =====================
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.

```

kode-ninja:
Hi,

Would docker work?

eng1neer:
Is `portia-orm-on-django` branch is obsolete now and `nui-develop` should be used instead?

ruairif:
`nui-develop` is also obsolete. We are now using `develop`. This should be released to `master` once we have the last few bugs fixed and the docs up to date.
@eng1neer How are you running portia at the moment? Using docker or vagrant?

eng1neer:
@ruairif I'm trying to run it via docker, but I can't figure out how to create a new project. In the old UI it was sufficient to navigate to a URL and the project would be created automatically. In the new UI there's only a blank screen with an empty list of projects. Is there an API call that can be used to create a project?

ruairif:
@eng1neer Could you try out the changes in #670 and see if that works for you

eng1neer:
@ruairif That worked perfectly, thank you!

PR #670, opened 2017-01-05
Add input box to create project

ruairif:
create_project
[Fix problem with project creation](https://github.com/scrapinghub/portia/issues/648#issuecomment-270632607)


michalmo:
Maybe this should be in the reverse order. Choose a project or create a new one.

ruairif:
when it's done this way both options are easily available without needing to scroll

eng1neer:
When creating a project with the name "example.com" it gets added to the project list but an alert flashes `Project with id 'undefined' not found`. I guess that dot symbol is not allowed in project name?

sagelliv:
@eng1neer Did it work with a non-dotted name then?

eng1neer:
@sagelliv  It did, but the project with dotted name was still created in somewhat corrupted state. It refused to load on click with the same error.

ruairif:
@eng1neer I've updated the PR to disallow project names containing a `.` thanks for pointing this out

PR #717, opened 2017-02-22
Only check project status if version controlled

ruairif:


portiaui/app/services/changes.js
```diff
@@ -19,6 +21,9 @@ export default Ember.Service.extend({
     }),
 
     _checkProjectChanges() {
+        if (!this.get('versionControlled')) {
```
sagelliv: Combine this into the guard clause below:
```
if (!project || !this.get('versionControlled')) { return false; }
```

PR #858, opened 2018-05-19
Update utils.py

AkhilNairAmey:
Change open encoding to `utf-8`


ruairif:
I was thinking that it was strange that there would be non ascii characters in the spiders file but I guess it's possible that they can appear in the urls. I think we always save the json files as utf-8 so this should be safe

